Issue #99, opened 2016-05-04
Need to add license information for the Code Point data


caprenter:
See: https://www.ordnancesurvey.co.uk/business-and-government/licensing/using-creating-data-with-os-products/os-opendata.html

caprenter:
Closed thanks to https://github.com/OpenDataServices/grantnav/commit/d7b624ebf92860a09384185733167ce37954887e

PR #100, opened 2016-05-04
[#99] Adds licence information about code-point data

caprenter:
We need to do this under the terms of the licence of the data
Adds info to the licence document and also to the footer.
Adds a test to see that it is displayed on the home page

### DIFF
--- a/LICENCE.rst
+++ b/LICENCE.rst
@@ -21,3 +21,13 @@ Copyright 2016 Open Data Services Co-operative Limited.
 
 
 
+Code-Point Open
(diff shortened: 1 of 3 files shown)


caprenter:
I think this is something that does need to be done. It seemed easier to do it rather than leave it in the queue, but I haven't actually previewed the change to the footer, so it could be contentious.

Can be left until the next iteration for consideration

kindly:
This needs to close due to the repo reshuffle but will be reopended again afterwards

PR #175, opened 2016-06-28
Add ALL THE DOWNLOAD BUTTONS

robredpath:


grantnav/frontend/views.py
```diff
@@ -303,7 +306,7 @@ def get_clear_all(request, context, json_query):
     except KeyError:
         json_query["query"]["bool"]["filter"] = copy.deepcopy(BASIC_FILTER)
         current_filter = json_query["query"]["bool"]["filter"]
-
```
robredpath: I have a filter in Sublime Text that strips trailing whitespace, hence this stuff in my PR

robredpath:
Core functionality is in and buttons exposed on all relevant pages. Currently outstanding issues are:

- Not sure how the CSV-ifier handles newlines. Probably badly. Can fix with some template filtering magic. 
- £ signs in JSON aren't properly encoded and probably don't want to be there
- No tests

robredpath:
@Bjwebb I'd appreciate your eyes on this - I've documented the issues that I already know about above, but I'm sure there's more!

PR #180, opened 2016-06-29
Adds a take down policy page

caprenter:
The 360Giving site now has a take down policy.
This adds a page that basically tells people to go there for the info
Adds a link in the footer similar to that on the 360Giving site

Also fixes up the css in those footer links to remove a list-style-type which is visible otherwise
Adds tests for those links in the footer

### DIFF
--- a/grantnav/frontend/static/css/main.css
+++ b/grantnav/frontend/static/css/main.css
@@ -86,6 +86,12 @@ a.anchor {
     padding: 0;
 }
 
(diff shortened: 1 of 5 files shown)


Bjwebb:
Would be good to have http://www.threesixtygiving.org/take-down-policy/ as a clickable link.

caprenter:
Ha - good point! Will fix.

Issue #62, opened 2016-04-07
Cookie Policy


caprenter:
GrantNav needs a policy about user privacy and should incorporate the EU Cookie directive requirements

Bjwebb:
GrantNav doesn't currently use any cookies, but this is likely to change with https://github.com/OpenDataServices/grantnav/issues/21

robredpath:
Made a note about cookies on #21 . +1 to a privacy policy as part of going live

robredpath:
Turns out we already have cookies. We should add cookie control!

caprenter:
I'm not sure why this is assigned to me.

I'll rename this to Cookie policy, create a new Privacy policy issue to separate them out.

This could be a solution: https://pypi.python.org/pypi/django-cookie-law/1.0.1
but I don't think I'm the person to implement it.

caprenter:
I've implemented and there is a pull request: https://github.com/OpenDataServices/grantnav/pull/176

PR #181, opened 2016-06-29
[#62] Adds a cookie acceptance banner to the application

Bjwebb:
I've force unmerged and recreated https://github.com/OpenDataServices/grantnav/pull/176 because I think it has files it shouldn't.

Original comment:
> Uses the django-cookie-law package to do this

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,16 +28,16 @@ Steps to installation:
 * Install Elastic search
 * Run the development server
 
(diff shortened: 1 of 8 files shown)


Bjwebb:
@caprenter I don't think you should have commited the static directory into this. I can go back and amend the commit if you want.

Issue #167, opened 2016-06-21
When multiple organisations share a charity number, facets show the name of the first one that the code encountered


robredpath:
Satisfies #163 

<img width="274" alt="screen shot 2016-06-21 at 15 29 22" src="https://cloud.githubusercontent.com/assets/641009/16233079/f644ab26-37c4-11e6-90ad-ea34067c6c1d.png">

robredpath:
Implementation: download the list of names from charity commission, use that for facets

robredpath:
I'll do this, I'll need @Bjwebb 's help wiring up the facets though

robredpath:
@ekoner did you create a charities commission data source at some point? Could you share that work with me - it'd be useful for this? 

I think @kindly is going to have to actually implement this though, I've not been able to get to it this week :/

ekoner:
@robredpath The May 2016 extract_charity is here: https://drive.google.com/file/d/0B_hfAR6BAnvYS1JtTS1kdFlFdVk/view?usp=sharing
You'll want to read the guidance here: http://data.charitycommission.gov.uk/data-definition.aspx

Charity Commission publish open data as bcp (SQL Server) files, there's a script for converting from NCVO which I've beefed up with some additional checks. Shout if you need it.

PR #185, opened 2016-07-07
[#167] Use charity registry

kindly:
Add script to download latest charity registry and convert it to json
lookup.  Use this primarily to look up charity names for facets.

Update requirments.


dataload/fetch_charity_data.py
```diff
@@ -0,0 +1,75 @@
...
+from io import StringIO
+import zipfile
+import os
+import csv
+import titlecase
+import json
+
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
+latest_zip_file = os.path.join(current_dir, 'charity_registory.zip')
```
robredpath: Typo - charity_registry.zip, presumably?

dataload/import_to_elasticsearch.py
```diff
@@ -165,6 +166,10 @@ def import_to_elasticsearch(files, clean):
 
     time.sleep(1)
 
+    with open(os.path.join(current_dir, 'charity_names.json')) as fd:
```
robredpath: What happens if this file is missing? Is there anything here to ensure that the script that grabs charity numbers is called at least once during a deploy? Or is that one for @Bjwebb ?

kindly: The actual json file is in the repo, so it should always work.  The script is just for when it needs updating

PR #190, opened 2016-07-08
169 stats search

kindly:


robredpath:
@kindly I kinda feel like the stats should appear above the 'results' header and the download buttons to make it clear that what you're downloading is a list of grants, not the stats. We've also had feedback that the mean isn't useful, so could you remove that? 

I've merged one of your other PRs and resolved the merge conflicts, so hopefully all that's left is your thoughts on ^^

kindly:
@robredpath I tried it both ways when implementing this but I preferred this way (very slightly) as I thought it odd that results count should be before the stats and that it looked better that heading "Results" was at the top not under the stats block.  I am not too bothered either way though.

robredpath:
Just to document the discussion: myself and @kindly discussed this after standup and played with some options. We're leaving it as it is, apart from removing the average and moving the results count into the summary box.

PR #218, opened 2016-07-19
[#216] Replace green/red icons with up down arrows in data tables

caprenter:
Uses CSS to change the icons used in the data tables on e.g. a funders page

Re-orders where 'extra-style' is loaded in the base template so that our
main.css file gets loaded last. This may have unintended consequences, please review.

### DIFF
--- a/grantnav/frontend/static/css/main.css
+++ b/grantnav/frontend/static/css/main.css
@@ -90,6 +90,32 @@ a.anchor {
   white-space: normal;
 }
 
+table.dataTable.dtr-inline.collapsed > tbody > tr > td:first-child::before, table.dataTable.dtr-inline.collapsed > tbody > tr > th:first-child::before {
+    top: 8px;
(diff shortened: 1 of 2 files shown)


robredpath:
I feel like the arrows should point right when closed, and down when open. See http://ux.stackexchange.com/questions/5735/which-way-should-arrows-point-in-a-collapsible-accordion-left-down-down-up for some discussion of this. 

Thoughts?

caprenter:
No thoughts for me. If you want to use a horizontal arrow the css would be \25b6 
(from https://brajeshwar.github.io/entities/ )

robredpath:
Does anyone else have an opinion on this? @Bjwebb @kindly ?

Bjwebb:
> I feel like the arrows should point right when closed, and down when open.

Yep, I'd agree with that.

caprenter:
I'll change this...

Issue #212, opened 2016-07-19
Rounded corner bug on Main banner


caprenter:
I'm seeing small black rounded corners on the main header in grant nav

PR #219, opened 2016-07-19
[#212] Rounded corner nav bug

caprenter:
Removes the rounded corner css on the navbar


caprenter:
Changes to theme.css should be made in the SASS files: https://github.com/OpenDataServices/grantnav#compile-theme

PR #233, opened 2016-07-27
207 search specific fields

kindly:


robredpath:
@kindly please can we move the 'new search' button? And/or rename it? Despite the fact that I know exactly what it does, I instinctively clicked the thing that had the word 'search' on it and wondered why my results were weird!

robredpath:
Actually, it's ok, I'll add it as an issue for next sprint :)

robredpath:
@kindly I'm not sure if bfefa6026c80a092eb53bc919bb4d480ef878dd5 is working as expected. When trying it out, the page still jumps to the relevant facet when I click 'see more'. Is there anything I might be missing?

kindly:
@robredpath it seems to work for me I will try another browser.

kindly:
@robredpath it did not work for chrome as it looks at fragments after it runs the javascript.  It is fixed but in chrome I hat the way it jumps to the top first but can not find a way to fix that.

PR #234, opened 2016-07-27
Publisher and "data used in GrantNav" pages

Bjwebb:
https://github.com/OpenDataServices/grantnav/issues/166

Demo at http://acceptable-license.166-publisher-pages.grantnav-dev.default.opendataservices.uk0.bigv.io/

Also includes substantial copyright related changes to the footer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ python manage.py runserver
 Note that if you are not on Debian based system you will need to follow https://www.elastic.co/guide/en/elasticsearch/reference/current/setup.html#setup-installation to install elasticsearch.
 Follow the instructions in your terminal to open the aplication in your browser.
 
+
+
 Upload Data
 ------------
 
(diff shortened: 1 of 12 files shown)


robredpath:
@Bjwebb when testing this, it looks like there's some data missing: 

screen shot 2016-08-08 at 18 40 32 1

I suspect this is due to my data, but when trying to load the data in, I get:

`(grantnav) [root@localhost grantnav]# python dataload/import_to_elasticsearch.py data/data_all.json
Updating existing index
data/data_all.json
Traceback (most recent call last):
  File "dataload/import_to_elasticsearch.py", line 356, in <module>
    import_to_elasticsearch(args.files, args.clean)
  File "dataload/import_to_elasticsearch.py", line 204, in import_to_elasticsearch
    keys = list(doc.keys())
AttributeError: 'list' object has no attribute 'keys'`

It works fine on the deployed copy of the branch. Can you confirm that this is just a data loading issue and that I haven't broken it while fixing merge conflicts?

Bjwebb:
The provenance json isn't meant to be loaded into elasticsearch, which is why that load fails.

I've deployed your rebased branch to our dev server - http://all.166-publisher-pages.grantnav-dev.default.opendataservices.uk0.bigv.io/ - which looks okay to me.

I suspect the problem you have locally is that the provenance lookup relies on the filenames data was loaded with.

PR #267, opened 2016-08-15
193 display unclear

kindly:


robredpath:
This certainly fixes the issue on iPads, however I'm finding the very heavy text a bit jarring. Not sure if that's just because I'm used to the old version though.

Bjwebb:
We said we'd merge this for now, and work out what we need to do for this going forward next week.

Issue #246, opened 2016-08-02
As a user arriving at the site, I want to know that my use case is catered for so that I feel confident using the tool


robredpath:
Carried over from #235, we should provide some clues on the front page that when users arrive with a certain use case in mind that it's supported

'I was also thinking this could possibly done in the form of user story quotes.
"I used grant nav to home in on some data then downloaded it..." Mrs Trellis etc!' - DC

robredpath:
@caprenter something to bear in mind when writing new examples for the front page

caprenter:
A way of doing this is explored in https://github.com/OpenDataServices/grantnav/pull/274

We actually need the user stories written up that we want to promote tho.

Issue #213, opened 2016-07-19
Better examples on front page


robredpath:
Quoted search, region, location, etc

caprenter:
I'm happy to have a go at something for this

caprenter:
Try: "arts centre", "Only locations: Bristol"

PR #274, opened 2016-08-16
[#213] [#246] An idea/option for explaining about GrantNav

caprenter:
Just adds some quotes to the front page. We can decide waht we want
them to say.

### DIFF
--- a/grantnav/frontend/static/css/main.css
+++ b/grantnav/frontend/static/css/main.css
@@ -230,3 +230,37 @@ table.dataTable.dtr-inline.collapsed > tbody > tr.parent > td:first-child::befor
     }
 }
 
+.quotes {
+  background: #fff;
+  margin-top:0;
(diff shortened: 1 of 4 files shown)


robredpath:
I've emailed Katherine at 360 to see what she thinks, as she's been thinking about contextless visits a lot. 

In principle, I like the idea of having something on the front page that people can instantly relate to and say 'yes, this is for me'. However, quotes like this are often used for reviews (so do they come across as too marketing-ey?) and it's a big design change.

caprenter:
This is probably really issue #246. I'll add that to the title of this issue.

caprenter:
Fine by me

robredpath:
As per our conversation earlier, I'm closing this for now. We might want to revive it in the future (potentially quite near future), depending on how the contextless visits conversation goes.

PR #276, opened 2016-08-18
[#213] Provides better exampls on the front page

caprenter:
Please review. It may not be quite waht is required.
Finding it hard to provide just the right kind of thing here.

### DIFF
--- a/grantnav/frontend/templates/home.html
+++ b/grantnav/frontend/templates/home.html
@@ -29,16 +29,12 @@
         </form>
       </div>
       <div class="col-xs-12">
-        <p class="search-help"> Try these examples: 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=volunteers">volunteers</a>, 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=heritage">heritage</a>, 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=disabilities">disabilities</a>,
-            
-            Only Locations: 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=Bristol&amp;default_field=recipientLocation">Bristol</a>,
-            Only Recipients: 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=children&amp;default_field=recipientOrganization.name">children</a></p>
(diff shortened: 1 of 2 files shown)


grantnav/frontend/templates/home.html
```diff
@@ -32,7 +32,13 @@
...
             <a class="link-on-orange" href="{% url 'search' %}?text_query=volunteers">volunteers</a>, 
             <a class="link-on-orange" href="{% url 'search' %}?text_query=heritage">heritage</a>, 
-            <a class="link-on-orange" href="{% url 'search' %}?text_query=disabilities">disabilities</a> 
+            <a class="link-on-orange" href="{% url 'search' %}?text_query=disabilities">disabilities</a>,
+            
+            Only Locations: 
+            <a class="link-on-orange" href="{% url 'search' %}?text_query=Bristol&amp;default_field=recipientLocation">Bristol</a>,
+            Only Recipients: 
+            <a class="link-on-orange" href="{% url 'search' %}?text_query=children&amp;default_field=recipientOrganization.name">children</a></p>
+        <p class="search-help">Tip: Use quotes to find an exact phrase <a class="link-on-orange" href="{% url 'search' %}?text_query=%22arts centre%22">"arts centre"</a> or identifier <a class="link-on-orange" href="{% url 'search' %}?text_query=%22360G-wolfson-19750%22">"360G-wolfson-19750"</a></p>
```
robredpath: I think this can be condensed to just an example in quotes at the end of the previous line - people will get the idea!

grantnav/frontend/templates/home.html
```diff
@@ -32,7 +32,13 @@
         <p class="search-help"> Try these examples: 
             <a class="link-on-orange" href="{% url 'search' %}?text_query=volunteers">volunteers</a>,
```
robredpath: We can probably remove at least one of these examples, if not two

caprenter:
I think this is a case of preferences. It seems to me as though @robredpath has stronger opinions, so I'm going to assign to him to edit as he sees fit.

robredpath:
I'm not sure that 'he who has the strongest opinions dicates the route' is the best way, but I'm happy to update these.

robredpath:
Updated @caprenter , what do you think?

caprenter:
This is fine by me

PR #285, opened 2016-08-19
284 stream upload

kindly:
#284

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ django-cookie-law==1.0.8
 titlecase==0.8.1
 dealer==2.0.5
 django-super-favicon==0.6.0
+ijson==2.3
 ## The following requirements were added by pip freeze:
 contextlib2==0.5.3
 django-classy-tags==0.7.2
--- a/dataload/import_to_elasticsearch.py
+++ b/dataload/import_to_elasticsearch.py
@@ -11,6 +11,8 @@ from pprint import pprint
 import elasticsearch.helpers
 import requests
 import time
+import ijson
+
 
 ES_INDEX = os.environ.get("ES_INDEX", "threesixtygiving")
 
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,22 +1,24 @@
...
 django-bootstrap3==7.0.1
-requests==2.10.0
+requests==2.11.1
 django-environ==0.4.0
-raven==5.23.0
-elasticsearch==2.3.0
--e git+https://github.com/OpenDataServices/flatten-tool.git@dceac82323271c4cc5956275481d762d81976eaf#egg=flattentool
+raven==5.24.3
+elasticsearch==2.4.0
+-e git+https://github.com/OpenDataServices/flatten-tool.git@dceac82323271c4cc5956275481d762d81976eaf#egg=flattentool-dev
```
robredpath: Is this some reqs weirdness?

Bjwebb: Yes, it's updated everything to the latest version. Reverting the two .txt files and running `./update_requirements --new-only` should fix this https://github.com/OpenDataServices/grantnav#adding-and-updating-requirements (or we could choose to have requirements updated)

Bjwebb:
This still uses ~ 1 GB of memory for me.

I wonder if we should be using `streaming_bulk` instead? http://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.streaming_bulk

Also, now that we have a [package schema](http://www.threesixtygiving.org/standard/reference/#toc-360giving-json-schemas) it's probably safer now to assume `grants_key = "grants"` instead of trying to guess it.

Bjwebb:
Although, checking a data load with the previous code, that gets closer to 2GB, so this could still be worth merging.

Bjwebb:
I've modified the grants key behaviour to always assume "grants". @kindly could you check that looks alright?

I'm happy for this to get merged, because it does seem to be an improvement.

Bjwebb:
> I wonder if we should be using streaming_bulk instead? http://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.streaming_bulk

Looks like `streaming` calls `streaming_bulk` under the hood, so this isn't relevant.

PR #286, opened 2016-08-19
Add favicons

robredpath:


requirements.txt
```diff
@@ -9,14 +9,32 @@ python-dateutil==2.5.3
 strict-rfc3339==0.7
 django-cookie-law==1.0.8
 titlecase==0.8.1
+django-super-favicon==0.6.0
 ## The following requirements were added by pip freeze:
 contextlib2==0.5.3
+coverage==4.1
```
Bjwebb: Some of these are dev requirements, and shouldn't be in this file.

robredpath: Hmm. I just ran update-requirements.sh --new-only and committed the result. Then again, I think that script relies on virtualenv and I'm running pyvenv, which may mess with things.

robredpath:
@Bjwebb fixed reqs, I think. Can you check?

Bjwebb:
The dev requirements in the wrong place is fixed.

This still updates all requirements, which I'm not sure if we're wanting to do. update_requirements has a --new-only flag if you want to avoid this https://github.com/OpenDataServices/grantnav#adding-and-updating-requirements

robredpath:
I ran it with --new-only but I did have to do some slightly brutal things to get it to work with pyvenv which may well have messed with its normal operation. 

Is updating an issue if it passes tests?

Bjwebb:
Updating everything's fine to do, but it will break other pulll requests, I think.

robredpath:
Tbh I'm manually fixing merge conflicts on most of them anyway so one more file won't make much difference!

robredpath:
@Bjwebb @kindly can one of you merge this, if you're happy? I'm happy to take the consequences of merging in reqs for the other PRs

Bjwebb:
I've changed my mind a bit on this - I think keeping the requirements change separated out will help us if we need to revert it later. I'll have a go at fixing and rebasing the relevant pull requests shortly.

Bjwebb:
Chrome loads 10 of these favicons, I'm not entirely sure if that's the behaviour we want..

Bjwebb:
We think this is a chrome bug, although one to keep an eye on if we have page load issues.

Issue #278, opened 2016-08-18
Link from funder page to publisher page


robredpath:
Eg http://all.master.grantnav-dev.default.opendataservices.uk0.bigv.io/funder/GB-CHC-200051

Add at top, under name/key stats: 

This data was originally published by Esmee Fairbairn Foundation see their GrantNav publisher page for more information.

robredpath:
![uploads/a2842489-9e5b-4f55-b0f9-a8da115a460d/Screen Shot 2016-08-18 at 16.59.14.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fa2842489-9e5b-4f55-b0f9-a8da115a460d%2FScreen+Shot+2016-08-18+at+16.59.14.png)

PR #287, opened 2016-08-19
[#278] Add link from funder to publisher

Bjwebb:


robredpath:
@Bjwebb I'm struggling to get this working. I've loaded the sample data set with import_to_elasticsearch.py and am running the server with PROVENANCE_JSON=data/data_valid.json python manage.py runserver 0.0.0.0:8008 , but I still get:
screen shot 2016-08-22 at 08 06 08

Bjwebb:
The lookup is happening based on the filename, so I think you need to make sure you're doing the load from the directory with the data in. ie, cd to it, so the relative path of all the files is just the filename. https://github.com/OpenDataServices/opendataservices-deploy/blob/master/salt/grantnav/reload_data.sh#L17

Bjwebb:
I've tried to update the README to make this clearer https://github.com/OpenDataServices/grantnav/tree/278-funder-publisher#provenance-json

PR #299, opened 2016-08-30
Added dropdown, updated CSS

notshi:
Minor comments added for referencing, radio button section needs deletion if all looks ok.


kindly:
This looks a bit dodgy in firefox at the moment.

firefox-grantnav-home

notshi:
What size screen are you viewing this in? I could only replicate this when my current screen is smaller than full size which makes the CSS media query for .container @min-width: 992px to kick in.

1. We could make the font smaller for the dropdown - 20px
2. The dropdown arrow which has a grey background is down to Firefox (browser dependant) which can be changed but will include unneccasary hacks (using a background image, etc).
3. The search examples could be fixed with any of these:
  - decrease number of examples to fit width
  - decrease section width so the examples flow better to two lines
  - decrease font size
  - replace spaces with non-breaking spaces (&nbsp) to stop it splitting words

kindly:
I think my screen was below that media width as well at the time.  My biggest worry was the dropdown looking a bit bad in firefox but I noticed the other errors too when doing the screenshot.

1. I wonder if we should actually change the text, Maybe to just "search all", "locations", "recipients". or a smaller font size.

2. I personally think something is needed here.

3. I would think for this trying to use inline blocks and floats so that a whole example will go on the next line. Maybe just using &nbsp will work too.

Overall not too worried about it looking "not great" on smaller screens as long as its usable and does not look to bad at about ipad size.

notshi:
We can focus on iPad size (768 x 1024). Screenshots using browser in iPad mode with styles below applied.

1. Text shortened, font size remains unchanged at 25px
3. display:inline-block; for example links 

Firefox - noticed that the dropdown does not appear to have the grey background or the arrow
screen shot 2016-08-31 at 15 34 44

Chrome
screenshot_2016-08-31_15-29-21

Will look into fixing the grey background for the dropdown.

notshi:
Updated styles

1. Text shortened for first option only, font size remains
2. Custom arrow added for Firefox, resizes accordingly to screen size, *slight hack* please check if okay
3. display:inline-block; for example links

kindly:
Just looked at changes and it all seems good.  The only issue left is on ipad sized screens when "only locations" is selected you can not see the end it.  

So either just change it to "locations" or make the font smaller generally or make the font smaller with media query.

Think I am happy with the css hack, css is a hack anyway.

notshi:
Updated styles

1. Made font smaller with media query, text remains the same to be consistent with search examples

kindly:
I think the font needs to be even smaller 15px not 20px.  Actually testing this on an ipad now and it still looks like its overlapping.

Like the mobile fix.

notshi:
Updated to 15px.

Thanks! Don't have an iPad here so appreciate the feedback and using the browser for iPad simulation seems not as reliable as it could be.

Issue #307, opened 2016-09-01
JSON download failing due to circular ref


kindly:
Think this is due to provenance information having the circular reference.

kindly:
Pretty sure the below will fix it but unable to test.  i.e the dataset has a publisher which has datasets that have ab publisher which has datasets....

```
diff --git a/grantnav/provenance.py b/grantnav/provenance.py
index 92ba1a9..581ddba 100644
--- a/grantnav/provenance.py
+++ b/grantnav/provenance.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from collections import OrderedDict
 import json
+import copy
 
 by_publisher = OrderedDict()
 by_identifier = {}
@@ -14,7 +15,7 @@ if settings.PROVENANCE_JSON:
     for dataset in datasets:
         prefix = dataset['publisher']['prefix']
         if prefix not in by_publisher:
-            by_publisher[prefix] = dataset['publisher']
+            by_publisher[prefix] = copy.deepcopy(dataset['publisher'])
             by_publisher[prefix]['datasets'] = []
         by_publisher[prefix]['datasets'].append(dataset)
         by_identifier[dataset.get('identifier')] = dataset
```

Bjwebb:
@kindly Yes that patch fixes it, thanks. I'll put a commit together for this, along with some tests to avoid it happening again.

Issue #306, opened 2016-08-31
Recipient location headers in CSV don't match the standard


Bjwebb:
These fields are populated:
* Recipient Region
* Recipient District
* Recipient Ward

These are not fields in the standard.

The fields in the standard are in the CSV, but generally not populated:
* Recipient Org:Location:0:Geographic Code Type
* Recipient Org:Location:0:Geographic Code
* Recipient Org:Location:0:Name
* Recipient Org:Location:1:Geographic Code Type
* Recipient Org:Location:1:Geographic Code
* Recipient Org:Location:1:Name
* Recipient Org:Location:2:Geographic Code Type
* Recipient Org:Location:2:Geographic Code
* Recipient Org:Location:2:Name

robredpath:
We're going to rename the extra fields to make it clear that they're augmented data, not original data.

robredpath:
Note: do the same for date/time downloaded and license info

PR #308, opened 2016-09-01
[#306][#307] Modifiy CSV headings and fields

kindly:
Change order of fields so that grantnav added fields are at the end.
Add beneficiary fields.


Bjwebb:
I don't think this is enough beneficiary fields, there is a grant in the Macc data with 8.

kindly:
oops miscounted. tested with the macc data too.

Issue #292, opened 2016-08-23
Prepare a static holding page and procedure in case of Really Bad Things happening


robredpath:
Although we have a resilient architecture, having a static 'oops, things went really wrong' page that we can redirect to during an outage would be really useful

robredpath:
@caprenter has written the page, @Bjwebb is thinking about the procedure

PR #313, opened 2016-09-01
[#292] Adds a failsafe backup static html file

caprenter:
Does this start to get us towards a fix for #292


robredpath:
Is the procedure here to just rename fail.html to index.html if everything's on fire? Does the httpd conf we have on these boxes let us do that?

robredpath:
I'm happy with the contents of the file, @Bjwebb is working on procedures for its use

Issue #301, opened 2016-08-31
'Developers' link in footer


robredpath:
Add static page, link in footer
Static page is for API links and welcoming pull requests

PR #316, opened 2016-09-01
[#301] [#302] Devs page and footer links

robredpath:


robredpath:
Thoughts on this, @Bjwebb @kindly @caprenter ?

Bjwebb:
Looking good.

I think this page also should include the specific JSON keys that we have added to the data. I'm happy to do that if you want.

> GrantNav is open source software by 360Giving.

Might be worth saying "developed by Open Data Services for 360Giving". Much like the T&Cs page http://grantnav.threesixtygiving.org/terms

> The data is updated at most once a day (except in case of an error), and usually once per week if there is new data available for us to load.

Possibly useful to link to http://grantnav.threesixtygiving.org/datasets/#datasets-used for the specific dates we last captured the data.

> ```<a href="htttp://grantnav.threesixtygiving.org/api/grants.json">```

typo: `htttp`

robredpath:
Made the tweaks you've suggested @Bjwebb , but if you could add the JSON keys that would be awesome

Bjwebb:
New issue about documenting JSON keys https://github.com/OpenDataServices/grantnav/issues/326

Will merge the rest of this shortly.

PR #321, opened 2016-09-02
[#307] Fix circular ref in JSON output

Bjwebb:
#307 

Also includes tests of this, which means wiring up the tests to load the provenance data.


robredpath:
@Bjwebb testing this is just checking that JSON downloading works again, right?

Bjwebb:
That's correct. @kindly may also be interested in how the tests have been changed, but that doesn't stop this getting merged.

Issue #318, opened 2016-09-02
BOM missing


Bjwebb:
I think we need this to cajole excel into opening UTF-8 files correctly.

PR #323, opened 2016-09-05
[#318] Add UTF-8 bom to CSV output

Bjwebb:
This helps spreadsheet programs realise that it's UTF-8.


Bjwebb:
Would be useful for someone to check this with a copy of MS Excel.

robredpath:
screen shot 2016-09-05 at 10 54 33

robredpath:
@Bjwebb is ^^ correct?

Bjwebb:
Looks good. Could you double check that it looks mangled when downloaded from the live site?

robredpath:
screen shot 2016-09-05 at 11 32 41

PR #363, opened 2017-01-04
Update Pillow dependency due to security bug

Bjwebb:


Bjwebb:
I'm going to test deploying this on dev before merging.

Bjwebb:
We only use Pillow for deploying favicons. I've tested on the dev server that the favicons still deploy fine.

PR #365, opened 2017-01-10
State that the license doesn't apply to the theme

Bjwebb:
Modelled after the note in https://github.com/OpenDataServices/cove/blob/master/LICENCE.rst

### DIFF
--- a/LICENCE.rst
+++ b/LICENCE.rst
@@ -1,8 +1,9 @@
 GrantNav
 ========
 
-Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png
-The look-and-feel is that of 360Giving, and care should be taken when running any public instance not to falsely give the impression, intentionally or otherwise, that the instance is operated by or related to 360Giving.
+Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png . The 360Giving logo is copyright `360Giving <http://threesixtygiving.org>` and may not be used without express permission. 
+
+The look-and-feel of GrantNav is that of 360Giving, and care should be taken when running any public instance not to falsely give the impression, intentionally or otherwise, that the instance is operated by or related to 360Giving.
 
 GrantNav is free software designed to help people explore data 
 published to the 360Giving open data Standard.
@@ -20,7 +21,7 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with GrantNav.  If not, see <http://www.gnu.org/licenses/>.
 
-Copyright 2016 Open Data Services Co-operative Limited.
+Copyright 2016-2017 Open Data Services Co-operative Limited.
 
 
 
--- a/LICENCE.rst
+++ b/LICENCE.rst
@@ -4,7 +4,7 @@ GrantNav
 Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png
 The look-and-feel is that of 360Giving, and care should be taken when running any public instance not to falsely give the impression, intentionally or otherwise, that the instance is operated by or related to 360Giving.
 
-GrantNav is free software designed to help people check data 
+GrantNav is free software designed to help people explore data 
 published to the 360Giving open data Standard.
   
 GrantNav is free software: you can redistribute it and/or modify
--- a/LICENCE.rst
+++ b/LICENCE.rst
@@ -1,8 +1,10 @@
 GrantNav
 ========
 
+Note: This license doesn't apply to the 360Giving theme and related assets (such as the logo).
+
 GrantNav is free software designed to help people check data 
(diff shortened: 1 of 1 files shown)


LICENCE.rst
```diff
@@ -1,8 +1,10 @@
 GrantNav
 ========
 
+Note: This license doesn't apply to the 360Giving theme and related assets (such as the logo).
+
 GrantNav is free software designed to help people check data 
-pubplished to the Open Contracting Data Standard.
+pubplished to the 360Giving open data Standard.
```
robredpath: published*

LICENCE.rst
```diff
@@ -1,8 +1,10 @@
 GrantNav
 ========
 
+Note: This license doesn't apply to the 360Giving theme and related assets (such as the logo).
```
robredpath: My feeling is that 'related' leaves too much to interpretation, especially as this file doesn't describe the license that is attached to those works. 

Are there any works in this repo that aren't covered by this license? If there aren't, that could be a neat way of distinguishing.

Bjwebb: My intent was to cover everything that we copied from the 360Giving website, as we don't have permission to release it under a FOSS license.

The next question then is what of that is copyright-able?

I think the logo is:
https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png
https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png

Beyond that I was worried about the "look and feel" of the theme. That the choice of colours / font / layout might constitute a copyright-able artistic work. I've done some brief googling, and I'm not sure if that is true.

robredpath: I believe that, in the UK at least, look-and-feel falls into trademark territory rather than copyrightability. The code to bring about that look-and-feel can be released under an open license, though - we don't care if someone downloads it, changes some numbers and makes a new blue CoVE. 360 would have something to say if they put something on the Internet that made out that it was related to them - but that would be under passing-off rules rather than copyright. 

Does it make sense for the license to cover all code but not logos, and then to provide a notice that the look-and-feel is that of 360Giving and care should be taken if running public copies of CoVE not to potentially confuse users?

Bjwebb: Yes, I think that makes sense.

Bjwebb:
@robredpath Could you check this updated (grantnav) pull requests. If this is okay, I'll make a similar update to the Cove one.

LICENCE.rst
```diff
@@ -1,7 +1,8 @@
 GrantNav
 ========
 
-Note: This license doesn't apply to the 360Giving theme and related assets (such as the logo).
+Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png
+The look-and-feel is that of 360Giving, and care should be taken when running public instance not to confuse users.
```
robredpath: "when running any public instance not to falsely give the impression, intentionally or otherwise, that the instance is operated by or related to 360Giving"?

Bjwebb: Good words, I'll use those.

Bjwebb:
I've updated again with the suggested text.

robredpath:
+1 from me. 

@stevieflow , is it appropriate to get 360Giving to review this as well, as it does relate to their IP, or is it just us as it's 'our' software?

stevieflow:
@robredpath @Bjwebb Im not exactly sure what/where to review

robredpath:
@stevieflow take a look at the 'files changed' tab - we're looking at changing the license text for GrantNav's code to better protect 360's brand

robredpath:
@Bjwebb I've slightly hijacked this with a one-word change that 360 requested while we were working on this. @stevieflow and I are talking re: the copyright situation with GrantNav!

LICENCE.rst
```diff
@@ -1,8 +1,11 @@
 GrantNav
 ========
 
-GrantNav is free software designed to help people check data 
-pubplished to the Open Contracting Data Standard.
+Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png
+The look-and-feel is that of 360Giving, and care should be taken when running any public instance not to falsely give the impression, intentionally or otherwise, that the instance is operated by or related to 360Giving.
+
+GrantNav is free software designed to help people explore data 
+published to the 360Giving open data Standard.
```
caprenter: The term used through the SSOT is "360Giving Data Standard".

This is not necessarily consistent on e.g. http://www.threesixtygiving.org/standard/ - but I think for now it helps if we stick to "360Giving Data Standard" unless/until we find out different.

LICENCE.rst
```diff
@@ -1,8 +1,11 @@
 GrantNav
 ========
 
-GrantNav is free software designed to help people check data 
-pubplished to the Open Contracting Data Standard.
+Note: This license doesn't apply to the 360Giving logo https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo.png https://github.com/OpenDataServices/grantnav/blob/master/grantnav/frontend/static/images/360logo_invert.png
```
caprenter: What license does apply to the logo then? Is it copyright 360Giving?

stevieflow: This looks like it still needs some attention @robredpath - regarding @caprenter question

Bjwebb: I don't know, I would assume All Rights Reserved (c) 360 Giving
I'm not sure what we want to put anything in our license file about this, and whether we should clarify it with 360Giving.

Bjwebb: @caprenter I don't know. We could just link to http://www.threesixtygiving.org/terms-conditions/ ?

PR #399, opened 2017-09-25
Updates README - takes out line about testing

stevieflow:
> This application is currently undergoing final testing before a production launch.

is no longer true


Bjwebb:
@robredpath could you quickly review, since I've also made my own changes here.

PR #404, opened 2017-10-18
Fix descriptions of arguments to import_to_elasticsearch.

stevage:


Bjwebb:
Thanks for taking the time to fix these!

Issue #414, opened 2017-12-08
Server Error (500) error related to Funder name


KDuerden:
I wanted to link to GrantNav with results filtered by funder Joseph Rowntree Foundation.
I couldn't find them in the Funder filters (they only show first 55, I think) so I tried to search for 'Joseph Rowntree Foundation' but got a Server Error (500) error. 
I get this when searching for Joseph Rowntree too, but not just Joseph. I tried other search terms 'mental health' 'domestic violence' are not affected. Joseph Rowntree Charitable Trust as a search term is not affected either!
When I did a search for York, I was able to filter by JRF. However I got the server error again if I run search without search terms with JRF filtered.

Bjwebb:
Looks like this is due to this issue: https://github.com/OpenDataServices/grantnav/issues/386

KDuerden:
So can be fixed via an update to the data - eg there is a stray character in there making it break?

Bjwebb:
Yes the problem is a newline in a recipient id `GB-CHC-\n1120420` (in the file "JRF York Committee’s small grants awarded in 2017"), so that can be fixed in the data.

Bjwebb:
I think there's a few other similar problems in that file also.

KDuerden:
I had seen those (eg few broken org Id links) so was planning to feedback. Hadn't realised there would be this knock-on effect. Thanks!

Bjwebb:
Looks like I can also fix the problem in GrantNav by allowing newlines in urls. The data still needs updating, as the identifier with/without the newline become different pages; but updating GrantNav would mean the failure isn't so broad (ie. breaking an entire search results page, when only a few results are wrong).

morchickit:
Can we close this issues? Or this still happens?

PR #417, opened 2017-12-12
[#414] Allow newlines in identifiers

Bjwebb:
#414


Bjwebb:
Actually, it seems that `(?s)^` doesn't work - I get a NoReverseMatch error like we did before.
I have added a comment though.

Issue #382, opened 2017-06-13
Download file - including the codes for admingeo areas


stevieflow:
In both the CSV and JSON download files, the relevant admin codes for each admingeo area are missing.  For example:

"recipientWardName":"Billesley Ward",
"recipientRegionName":"West Midlands",

The admingeo codes are unique - the names will not always be.  Secondly, the codes can be used by people to cross match with other areas

First of all - do we store the admingeo codes in GN?  It would be useful to know, to then understand if we can distribute them in the download files

stevieflow:
This case study (that @KathernDuerden pointed out) kind of supports this,,, nearly: http://www.threesixtygiving.org/2016/12/16/putting-local-charities-on-the-map/ !

robredpath:
What you download with the JSON or CSV download is everything we've got in GrantNav. 

The intention with providing regions and wards was to power the facets. GrantNav uses a few different strategies to work out where a grant is based on the geo data that's provided in the grant - for example, based on the recipient postcode if that's provided. Some publishers provide geocodes, and if they do, they're included in the download. 

From my quick look into it this morning, I believe that it would be possible to provide geocodes for more grants if we have some geo information to start with.

stevieflow:
@robredpath thanks - but you miss the point slightly :)

When we load to GrantNav it is true that we convert certain data into ward, local authority and region admingeo areas.  In the CSV file - for example - we get the following in columns BJ to BL (which is the data GN has added to the source)


```
Recipient Region	Recipient District	Recipient Ward
West Midlands	Birmingham District (B)	Nechells Ward
West Midlands	Birmingham District (B)	Weoley Ward
West Midlands	Birmingham District (B)	Moseley and Kings Heath Ward
West Midlands	Birmingham District (B)	Nechells Ward
West Midlands	Birmingham District (B)	Moseley and Kings Heath Ward

```
What we are asking for is what ID / code is also available at source, which could be included alongside the string for the name of each area.

robredpath:
@stevieflow sorry if I wasn't clear - if we have enough information to add those columns, then we have enough information to add geocodes for the columns that we currently have strings for. The codelist at https://github.com/OpenDataServices/grantnav/blob/03ebd5ed6e6bccb50686bfa3b264b29a9ee425e0/dataload/codelist.csv has everything we need to do that, so it's just a case of writing the code to do it!

KDuerden:
I love this file! Means I can at least do lookups in the interim 😊

[...]

Issue #345, opened 2016-10-03
Some admin geography areas have a (B) in their name


stevieflow:
I noticed a few geo areas that include a `(B)` after them: 

http://grantnav.threesixtygiving.org/district/Doncaster%20District%20(B)

Is there a reason for this?

stevieflow:
ping

robredpath:
Looks like the (B) stands for Borough

robredpath:
...and it's that way because that's what's in CodePoint Open. 

Labelled with DWI, we can attend to this next time we're devoting development resources to GrantNav

robredpath:
This is in the loading script. We know how to do this.

edugomez:
@robredpath @stevieflow Do we want to get rid of the _'(B)'_ s or replace them with _'Borough'_? Is not very clear in the ticket.

robredpath:
@edugomez good point. I'd assumed we were removing the (B) but actually replacing with (Borough) could be useful for people.

stevieflow:
Would the "Borough" end up in some long strings, and potentail line breaks/wrap?  I'd just lose the (B)

robredpath:
@stevieflow there are already a couple of names where the lines wrap. But, if you don't think that it's useful for people (and, on reflection, using the (Borough) meaningfully feels like analysis, which is beyond the scope of GN), then we might as well remove it. 

@edugomez - let's go with removing the (B) entirely. If it's the wrong thing to do we can always change it!

robredpath:
We removed the (B)

PR #431, opened 2018-01-12
Include geo codes in data added by grantnav

edugomez:
#382 

This PR also gets rid of _(B)_ s in district names as for #345

### DIFF
--- a/dataload/import_to_elasticsearch.py
+++ b/dataload/import_to_elasticsearch.py
@@ -23,6 +23,9 @@ bad_org_ids = []
 postcode_to_area = {}
 district_code_to_area = {}
 ward_code_to_area = {}
+district_name_to_code = {}
+ward_name_to_code = {}
+
 current_dir = os.path.dirname(os.path.realpath(__file__))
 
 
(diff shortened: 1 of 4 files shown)


robredpath:
@edugomez am I right in thinking that a geocode added by GN will be indistinguishable from one provided by the publisher?

edugomez:
@robredpath Yes

robredpath:
@morchickit @KDuerden can I just check that this ^^ you're ok with? It does mean that a user doesn't know if the geocode data has been added by GN, which might weaken the case for it being supplied by the publisher (but does also make the data more useful, and thus potentially more used....)

edugomez:
>  It does mean that a user doesn't know if the geocode data has been added by GN

The code in this PR only adds the geocodes supplemented by GN to the CSV/JSON downloads. We could do the following:
  
gn_geoc

In which case it is clear that the geographical data was added by GN (as the grant only includes post code). However, when the grant includes geo codes, users may get the impression that GN is duplicating data rather than enriching it.

Alternatively, it should also be possible to check whether the grant includes geo codes so as not show them under additional data in those cases.

robredpath:
Great, thanks @KDuerden

edugomez:
I have updated the code in this PR and now displays the section _Additional data added by Grantnav_ as in the screenshot above.

Issue #356, opened 2016-11-16
Preview text when sharing on social media is the cookie warning


robredpath:

robredpath:
![screen shot 2017-06-21 at 11 55 38](https://user-images.githubusercontent.com/641009/27380827-9522af18-5678-11e7-8d66-466b61258a20.png)

morchickit:
Shouldn't this be closed? @robredpath @BibianaC

robredpath:
Just tested on FB, and looks good now

PR #436, opened 2018-01-17
Add meta property="og:title" and property="og:description"

edugomez:
#356


edugomez:
@Bjwebb Any chance you could pull this branch to a live server instance? I think it should work but can't really check without a server connected to the internet.

Bjwebb:
Here's a deploy http://current.acceptable-license-valid.356-shared-url-shows-cookie-po.grantnav-dev.default.threesixtygiving.uk0.bigv.io/

edugomez:
@Bjwebb Thanks.

It seems to work for example in Hotmail web client:

hotmail

It should work with FB and other social media

PR #437, opened 2018-01-18
380 currency facet

kindly:


robredpath:
@kindly please can you switch the facet to displaying currencies in upper case?

kindly:
@robredpath they should display in upper case.  

Well actually they now "get loaded" in upper case and be upper case everywhere.   So basically we need to reload data before this works.

This is the better solution then just upper casing them in the display so that an input of "usd" and "USD" are equivalent and only show up once is the facet.

Bjwebb:
These commits are included in this deploy: http://current.acceptable-license-valid.420-other-currency-changes.grantnav-dev.default.threesixtygiving.uk0.bigv.io/
The load was rerun for this branch, so the currencies appear in upper case.

PR #438, opened 2018-01-18
420 funder currencies

kindly:
Funder and recipient summaries break down by currency.


robredpath:
@Bjwebb any chance you could load this onto the dev server so we can have a play?

Bjwebb:
I think these changes are included in this branch/deploy http://current.acceptable-license-valid.420-other-currency-changes.grantnav-dev.default.threesixtygiving.uk0.bigv.io/

PR #439, opened 2018-01-19
420 other currency changes

kindly:


grantnav/frontend/templates/currency_stats_modal.html
```diff
@@ -0,0 +1,27 @@
...
+      </div>
+      <div class="modal-body">
+        {% for bucket in results.aggregations.currency_stats.buckets %}
+          {% if not forloop.first %}
+            {% with currency=bucket|get:"key"|upper %}
+              <dl class="dl-horizontal top-stats-search">
+                <dt>Total {{currency}} grants</dt> <dd>{{bucket|get:"amount_stats"|get:"count"|get_amount}}</dd>
+                <dt>Total {{currency}} awarded</dt> <dd>{{currency|currency_symbol}}{{bucket|get:"amount_stats"|get:"sum"|get_amount}}</dd>
+                <dt>Largest {{currency}} award</dt> <dd>{{currency|currency_symbol}}{{bucket|get:"amount_stats"|get:"max"|get_amount}}</dd>
+                <dt>Smallest {{currency}} award</dt> <dd>{{currency|currency_symbol}}{{bucket|get:"amount_stats"|get:"min"|get_amount}}</dd>
```
edugomez: Given that ```get``` filter defaults to ```None```, I think these chains of ```get```s might _in principle_ give a 500 (_" 'NoneType' object has no attribute 'get' "_)

grantnav/frontend/templates/currency_stats.html
```diff
@@ -0,0 +1,25 @@
+{% load frontend %}
+
+<dl class="dl-horizontal top-stats-search">
+  {% with currency=results.aggregations.currency_stats.buckets|first|get:"key"|upper buckets=results.aggregations.currency_stats.buckets%}
+    <dt>Total {{currency}} grants</dt> <dd>{{buckets|first|get:"amount_stats"|get:"count"|get_amount}}</dd>
+    <dt>Total {{currency}} awarded</dt> <dd>{{currency|currency_symbol}}{{buckets|first|get:"amount_stats"|get:"sum"|get_amount}}</dd>
+    <dt>Largest {{currency}} award</dt> <dd>{{currency|currency_symbol}}{{buckets|first|get:"amount_stats"|get:"max"|get_amount}}</dd>
+    <dt>Smallest {{currency}} award</dt> <dd>{{currency|currency_symbol}}{{buckets|first|get:"amount_stats"|get:"min"|get_amount}}</dd>
```
edugomez: Given that ```get``` filter defaults to ```None```, I think these chains of ```get```s might in principle give a 500 (_" 'NoneType' object has no attribute 'get' "_)

grantnav/frontend/templates/currency_stats_modal.html
edugomez: Same than above for ```get```

grantnav/frontend/templates/currency_stats.html
kindly: If this was data that we got from a user then I would agree.  However this data is returned from elasticsearch so I believe we can rely on the structure being the same and all keys being there.

edugomez: :+1: I thought about that too, but trying to access a property of a ```None``` type is such a common error that I thought it was worth mentioning

robredpath:
@morchickit - it'd be great if you could take a look at these changes and +1 if you're happy for us to merge them and send them live. 

I *think* the commit messages are descriptive enough, and you can try them out at http://current.acceptable-license-valid.420-other-currency-changes.grantnav-dev.default.threesixtygiving.uk0.bigv.io/ - but let me know if you need any more detail I can provide it!

morchickit:
+1 from me. Let's roll it out

Issue #430, opened 2018-01-12
updating codepoint lookup and codelist


BobHarper1:
The [code list](https://github.com/OpenDataServices/grantnav/blob/iteration08/dataload/codelist.csv) and [codepoint open lookup](https://github.com/OpenDataServices/grantnav/blob/iteration08/dataload/codepoint_with_heading.csv.gz) in master and the public site iteration haven't been updated in a while (August 2016).

The effect within Grantnav would likely be neglible, but postcodes in the CodePoint database do get added to and changed frequently. The administrative area codelists also get updated more often than you would presume.

Here are two Ward names and codes in the current CodePoint
Clydebank Central Ward | S13003126
Clydebank Waterfront Ward | S13003127

They appear to have changed since the last codelist.csv was created: https://github.com/OpenDataServices/grantnav/blob/master/dataload/codelist.csv#L9064-L9065

If admin codes are outdated then they won't match as expected for external users of the data (say I'm downloading off GN to draw up a GIS map with the latest boundary packs).

Is it worth submitting a PR with the latest codepoint and codelist?

kindly:
@BobHarper1 yes. 
We aimed to do this anyway when we were trying to add propper region code but a separate pull request would be better.

The instructions are here, if you had not seen:
https://github.com/OpenDataServices/grantnav#fetching-external-datasets

It is sadly hard to fully automate this because as far as I know you can not get the Code Point Open through a reliable and updated URL.

PR #440, opened 2018-01-19
update to Codepoint vOctober 2017

BobHarper1:
#430


Bjwebb:
This is now being used on the main dev site http://current.acceptable-license-valid.master.grantnav-dev.default.threesixtygiving.uk0.bigv.io/

Issue #394, opened 2017-08-24
Fix text on developers page


morchickit:
The whole data set is over 100MB and growing.. .
I just downloaded the new json file and it's way over 300MB (331MB to be exact). We need to signpost it (that it's now 331MB) better and explain how people can get slices of the data (basically by searching grantnav).

@stevieflow and @robredpath - we discussed about it last week, but now I actually downloaded the data, so can give better figures.

morchickit:
Let me correct- the JSON is 331 MB, the CSV is 172MB. We just need to make sure people aware that the JSON is a monster

stevieflow:
@morchickit thanks.  Please can you suggest the text change we need (which will be (nearly) futureproof):

Currently - this says: http://grantnav.threesixtygiving.org/developers

> The whole data set is over 100MB and growing, so we recommend that you store the data locally for your own application to work from.

robredpath:
Current text: 

` The whole data set is over 100MB and growing, so we recommend that you store the data locally for your own application to work from. The data is updated at most once a day (except in case of an error), and usually once per week if there is new data available to load. A list of all data sets and the date on which they were last updated can be found at http://grantnav.threesixtygiving.org/datasets/#datasets-used`

How about this for new text, @stevieflow @morchickit ?

`The current data set is several hundred MB, and growing, so if you are using the data in an application we recommend that you download it and store it locally for your own application to work from. The data is updated at most once a day (except in case of an error), and usually once per month. A list of all data sets and the date on which they were last updated can be found at http://grantnav.threesixtygiving.org/datasets/#datasets-used

If you are downloading the whole data set for analysis, you may find that the size makes it very difficult or impossible to handle with desktop applications. We recommend that you carry out initial filtering or segmentation of the data with the GrantNav search interface and use the download links on the search results page to download the results in either JSON or CSV format. `

stevieflow:
looks good

> The data is updated at most once a day

This might confuse - as we mainly update GrantNav once a month.  We're talking about GrantNav data here?

robredpath:
@stevieflow hence why that sentence goes on to say "and usually once per month". Would "The data is typically updated once a month, and at most once a day ..." be clearer? The intention of that sentence is to discourage people from downloading the whole file more than once a day to check for changes.

stevieflow:
@robredpath tbh, If we are talking about GN, then I'd just stick to the once a month story - as per: 

http://grantnav.threesixtygiving.org/datasets/

[...]

PR #445, opened 2018-02-05
[#394] Update size of data set on devs page

robredpath:


robredpath:
@morchickit @stevieflow can you check this change to the copy on the dev page, following discussion in #394 ?

robredpath:
This is ready to merge, when approved.

PR #455, opened 2018-03-08
the 360Giving Company Number in the footer is wrong

BobHarper1:
The zero is missing from the front of the number. Is this an in-joke?

Also, I added links to the respective pages on Co House and Charity Commission


andylolz:
Ooh – the same footer is on this page: http://www.threesixtygiving.org/support/company-numbers/

Issue #464, opened 2018-06-11
ward codes incorrect


drkane:
Haven't looked at why - but it looks like some of the values in "Recipient Ward Geographic Code" are incorrect. It seems to be that they are based on a lookup to the ward name rather than the code, which causes issue where wards have duplicate names.

For example, grants that should have been in ward `E05000400` - Alexandra Ward in Kingston Upon Thames were instead assigned to `E05000719` - Alexandra Ward in Oldham.

drkane:
Looking into it a bit further, I think [/dataload/import_to_elasticsearch.py#L388-L399](https://github.com/OpenDataServices/grantnav/blob/master/dataload/import_to_elasticsearch.py#L388-L399) is part of the problem here. 

I think adding `ward_code` to the `postcode_to_area` dict, and adding a check for `ward_code` to the `add_area_to_grant()` function ([#L285](https://github.com/OpenDataServices/grantnav/blob/master/dataload/import_to_elasticsearch.py#L285)) would probably do the trick.

Bjwebb:
Yes, I think that would work. @kindly Could you double check?

morchickit:
@robredpath - Can we close this?

PR #465, opened 2018-06-21
Swap ward names to ward codes for lookup

drkane:
Fix for issue #464. 

Tests pass, and I tested on a specific data file which creates at least one grant with the wrong Ward code, and the correct code was created: http://grantnav.threesixtygiving.org/grant/360G-ClothworkersFdn-200805204

Potential complications - are there any circumstances where the Ward Name attached to a grant recipient is used to lookup this field? If there are then those cases won't work any more.


Bjwebb:
BTW, this change is live on http://grantnav.threesixtygiving.org/ now

drkane:
Looks great - the grant I mention above is now correctly showing E05000408 as the ward code

PR #467, opened 2018-07-06
15540-37 'and' vs 'or' queries [WIP - do not merge]

BibianaC:
Problem:
When a user searches for 'music therapy' (without quotes), the results can include 'music' or 'therapy'.
When a user searches for 'music therapy' (with quotes), the results include both words.
User could use 'and' or 'or' without knowing the impact in the search.

Solution:
Display information messages in those cases. The messages are a proposal, you can make better recommendations.

Without quotes:
'If you use quotes around your search, the result will be more accurate.'
And:
'"And" requires each word to be found'
Or:
'"Or" requires one of the two words to be found'

screen shot 2018-07-06 at 12 22 02

### DIFF
--- a/grantnav/frontend/templates/search.html
+++ b/grantnav/frontend/templates/search.html
@@ -32,6 +32,13 @@
       </form>
     </div>
   </div>
+
+  {% if search_help %}
+  <div>
+    <p>{{ search_help }}</p>
+  </div>
+  {% endif %}
+
   {% if selected_facets %}
   <div class="row">
     <div class="col-xs-12">
--- a/grantnav/frontend/templates/search.html
+++ b/grantnav/frontend/templates/search.html
@@ -34,7 +34,7 @@
       {% if advanced_search_info %}
       <div class="col-xs-10">
         <p>
-          {{ advanced_search_info }}
+          Tip: {{ advanced_search_info }}
           <a href="http://grantnav.threesixtygiving.org/help#advanced_search" class="advanced_search">Advanced Search</a>
         </p>
       </div>
--- a/grantnav/frontend/templates/search.html
+++ b/grantnav/frontend/templates/search.html
@@ -30,16 +30,6 @@
           </div>
         </div>
       </form>
-
-      {% if advanced_search_info %}
-      <div class="col-xs-10">
-        <p>
-          Tip: {{ advanced_search_info }}
-          <a href="http://grantnav.threesixtygiving.org/help#advanced_search" class="advanced_search">Advanced Search</a>
-        </p>
-      </div>
-      {% endif %}
-
     </div>
   </div>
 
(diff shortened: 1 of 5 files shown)


grantnav/frontend/templates/search.html
```diff
@@ -30,8 +30,20 @@
           </div>
         </div>
       </form>
+
+      {% if advance_search_info %}
```
robredpath: It's 'advanced search' in all of these cases, not 'advance search'

robredpath:
Hi @morchickit ! @BibianaC 's work here introduces the framework for displaying messages to the user when their search might be improved by using an advanced search feature - specifically:
- if a search contains spaces, then it might benefit from quotes
- if a search contains 'or' or 'and' then it might produce unexpected results if the user doesn't realise that they're using a boolean operator

We should agree on the exact words to be displayed in each of these cases. I've made some initial suggestions below. If you'd prefer that we did this in another way (eg Google Doc) then please do let me know and we can get it set up. 

Spaces: `Tip: If you're looking for a specific phrase, put quotes around it to refine your search. e.g. "youth clubs"`
OR: `Tip: The OR keyword (not case-sensitive) means that results must have one of the words present. This is the default. If you're looking for a phrase that has the word 'or' in (e.g. "NYC or bust"), put quotes around it`
AND: `Tip: The AND keyword (not case-sensitive) means that results must have both words present. If you're looking for a phrase that has the word 'and' in it, put quotes around the phrase (e.g. "fees and costs").`

@BibianaC - can we look at some options for the presentation of the message? Can we see what a thin-lined grey box would look like (as per the summary box at the top of the search results)? Would it be hard to add some colour - eg a green box so as to alert users that nothing's wrong but they might be able to improve their search?

grantnav/frontend/templates/search.html
BibianaC: ups yes! Thanks! I have now fixed it.

BibianaC:
@robredpath sure, I can change it.

morchickit:
The text looks good to me. Let's proceed with that.

BibianaC:
@robredpath @morchickit Grey box with text in black / green.

screen shot 2018-07-09 at 15 57 55
screen shot 2018-07-09 at 15 55 20

morchickit:
You might want to move above the Results header, no?

BibianaC:
@morchickit Something like that?

screen shot 2018-07-09 at 16 46 08
screen shot 2018-07-09 at 16 54 11

BibianaC:
@morchickit which one?

robredpath:
@BibianaC @morchickit I prefer the full-width one. 

@BibianaC - can we make the words wrap more nicely (ie, not break across lines), and replace the 'Advanced Search' link with "For more tips, see Advanced Search"

BibianaC:
@robredpath the sentences are quite long. If I force them to be in one line and the user is checking grantnav in a very small screen, he/she won't be able to read anything.

@morchickit let me know if you also prefer the full-width one.

morchickit:
I do prefer the full width one

robredpath:
Great! @morchickit - are you happy for us to merge and deploy this?

morchickit:
Sure, let's go for it!

PR #469, opened 2018-07-11
367 add links to advanced search

BibianaC:
I have added links to the examples in `Advanced search` documentation.

Some of the links don't work, it adds extra quotes. Lines: 239, 258, 288, 296 and 297.


robredpath:
This looks good, @BibianaC . I'm not sure about the orange underline though - it works on the orange link text elsewhere but it looks a bit weird under red example text. Can we make it red?

robredpath:
(I've tested all the links btw and they all work as expected :) )

BibianaC:
Thanks @robredpath . I will check the underline.

BibianaC:
@robredpath the underline is fixed.

robredpath:
Nice, thanks @BibianaC . @morchickit , what do you think?

BibianaC:
@morchickit gentle reminder that we are waiting for your feedback.

morchickit:
Sorry, missed this in my inbox. Checking

morchickit:
My only concern is fuzzy search, because at the moment it doesn't work properly

morchickit:
Other than the fuzzy, that has nothing to do with this code, we need to just document it well, looks ok!

PR #476, opened 2018-08-03
Django security update

Bjwebb:


Bjwebb:
We discussed elsewhere that 1.11.16 is not yet released, even though it has a page https://docs.djangoproject.com/en/2.1/releases/1.11.16/

PR #502, opened 2019-01-31
Update insecure requirements

Bjwebb:
Also pins some extra required dependencies.


morchickit:
GrantNav has 500 server error, does this have anything to do with it?

Issue #503, opened 2019-03-05
Multiple publishers use the "Reference" non-standard field in different ways


robredpath:
The most recent GN load encountered problems because two publishers had provided a "Reference" column. One used only numbers, the other used strings. Because of elastic's auto-schema-recognition stuff and the number one coming first in the load, elastic didn't know how to parse the one that used strings.

PR #504, opened 2019-03-05
Add Reference field to elastic schema

robredpath:
See #503 

This is a massive hack. But I did it on live to get the most recent GN load working. Alternative ideas (other than telling publishers not to do that, or re-ordering the registry!) welcome.


Bjwebb:
This seems like a reasonable thing to do. Probably worth @kindly having a look also as he knows the elastic schema better.

PR #511, opened 2019-07-22
Lot of changes to support ES7

kindly:
* Make import work by changing field mapping using new types.
* Fix various elastic search queries to support new api.
* Clean some queries to not have any empty lists or objects in.
* Update dependancies.
* Fix code to use new elasticsearc output format.
* Comment out stats page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
-dist: trusty
+dist: bionic
 addons:
   chrome: stable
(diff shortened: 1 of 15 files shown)


michaelwood:
Going to close this pull request in favour of a single 'master-next-major-version' branch which will contain both our sets of changes and some of which relate to datastore things.

PR #515, opened 2019-09-19
Mw/es7 configurable index

michaelwood:
Make the elastic index that is in use dynamically configurable. This will help us to have no down time when switching between indexes whilst a new data load takes place. Related to work for https://github.com/ThreeSixtyGiving/datastore/issues/5


michaelwood:
Note this pull request is on top of #511 and should only be merged when we are ready to migrate to es7

PR #535, opened 2020-02-07
templates Update copy to reflect introduction of dataservices

michaelwood:
Also includes a few grammar and misc copy updates.

@KDuerden FYI


michaelwood:
> The inconsistent use of {% blocktrans %} or not is also in the source, so I assume that's not a problem at the moment?

Thanks. Yes, not sure what the history of the translation support is. However currently it is out of scope.

PR #554, opened 2020-08-13
Mw/fix issue 343

michaelwood:


robredpath:
I've got this running locally:
* It does add commas to the grants count
* The "smallest" column seems to have disappeared 

@michaelwood

michaelwood:
> I've got this running locally:
> 
> * It does add commas to the grants count
> * The "smallest" column seems to have disappeared
> 
> @michaelwood

Hmm I can't reproduce the smallest column issue you mentioned, could you double check that?

Screenshot from 2020-08-14 11-55-18

robredpath:
Interesting. 

The column appears to be in the HTML but isn't getting shown. It displays for a second while the page is rendering, then disappears. 

It's the same on master, though. I'm happy to put it down to local weirdness (doesn't work on my machine?) and approve this. 

Screenshot 2020-08-14 at 13 44 43

PR #555, opened 2020-08-14
New es7 install instructions

robredpath:


robredpath:
Gah, branch messiness

Issue #562, opened 2020-09-22
Don't allow any other errors past other than already exists


michaelwood:
https://github.com/ThreeSixtyGiving/grantnav/blob/b078dc7297ccb680af247c7dd585c4d4690bdc61/dataload/import_to_elasticsearch.py#L233

We should probably be aborting or raising an exception at this point to abort any shell script that this runs in.

michaelwood:
Now fixed and deployed.

michaelwood:
![](https://github.trello.services/images/mini-trello-icon.png) [Make sure load fails gracefully](https://trello.com/c/tUYKCTOu/118-make-sure-load-fails-gracefully)

PR #563, opened 2020-09-22
dataload: import_to_elasticsearch: If error from ES raise an exception

michaelwood:
This helps us to abort any further process as this will generate the
correct exit status and cause the shell script that this is usually
triggered from to abort too.

https://github.com/ThreeSixtyGiving/grantnav/issues/562

### DIFF
--- a/dataload/import_to_elasticsearch.py
+++ b/dataload/import_to_elasticsearch.py
@@ -228,8 +228,12 @@ def maybe_create_index(index_name=ES_INDEX):
 
     # Create it
     result = es.indices.create(index=index_name, body={"mappings": mappings, "settings": settings}, ignore=[400])
-    if 'error' in result and result['error']['reason'] == 'already exists':
-        print('Updating existing index')
+    if 'error' in result:
+        if 'already exists' in result['error']['reason']:
+            print('Updating existing index')
+        else:
+            pprint(result)
+            raise Exception("Creating index failed")
     else:
         pprint(result)
 


dataload/import_to_elasticsearch.py
```diff
@@ -228,8 +228,12 @@ def maybe_create_index(index_name=ES_INDEX):
 
     # Create it
     result = es.indices.create(index=index_name, body={"mappings": mappings, "settings": settings}, ignore=[400])
-    if 'error' in result and result['error']['reason'] == 'already exists':
-        print('Updating existing index')
+    if 'error' in result:
```
robredpath, line 231: @michaelwood does this also catch / should we also catch circumstances where we can't talk to ES, temporarily?

michaelwood: The elasticsearch python library via requests/urllib will raise an exception if something goes wrong at the lower layers. Whereas this is just to raise an exception if the http request were sent and received and what was sent back was a json document with the error. 

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/indices.py#L99

BibianaC (approved):

michaelwood:
I've updated the commit because I found that the check for "already exists" string never actually worked on the current ES as they've added the index name into the reason field thus breaking the "==". We wouldn't have seen any problem with this as it was just printing the values out anyway.

BibianaC (approved):

PR #577, opened 2020-11-04
571 new params

kindly:


grantnav/frontend/templates/search.html
```diff
@@ -115,15 +115,15 @@ <h3 class="indent"> Filter By <a id="filter_by_help_link" href="/help#filters"><
                 <div class="col-lg-5 col-md-6 no-padding-hor">
                   <div class="form-group form-group-sm">
                     <label for="min-amount">Smallest
-                    <input id="min-amount" name="min_amount" class="form-control" value="{{query.query.bool.filter.3.bool.should.range.amountAwarded.gte}}"/>
+                    <input id="min-amount" name="new_min_amount" class="form-control" value="{{query.query.bool.filter.3.bool.should.range.amountAwarded.gte}}"/>
```
michaelwood, line 118: Is there a reason for the `new_` prefix?

grantnav/frontend/tests.py
```diff
@@ -53,23 +53,6 @@ def test_search(provenance_dataload, client):
...
 
-    # click facet
-    wolfson_facet = response.context['results']['aggregations']['fundingOrganization']['buckets'][0]
-    assert wolfson_facet['doc_count'] == 7
-    assert wolfson_facet['key'] == '["Wolfson Foundation", "GB-CHC-1156077"]'
-    json_query = json.loads(urllib.parse.parse_qs(initial_response.url.split('?')[-1])['json_query'][0])
-
-    json_query['query']['bool']['filter'][0]['bool']['should'] = [{'term': {'fundingOrganization.id_and_name': '["Wolfson Foundation", "GB-CHC-1156077"]'}}]
-    assert json.loads(urllib.parse.parse_qs(wolfson_facet['url'].split('?')[-1])['json_query'][0]) == json_query
-
```
michaelwood, line 64: We may as well keep the other assert tests e.g. `doc_count`, `key` and only remove the `assert json.loads(...` part

grantnav/frontend/views.py
```diff
@@ -491,6 +491,185 @@ def add_advanced_search_information_in_context(context):
...
+    json_query = copy.deepcopy(BASIC_QUERY)
+    json_query["query"]["bool"]["must"]["query_string"]["query"] = request.GET.get('query', '*')
+    json_query["query"]["bool"]["must"]["query_string"]["default_field"] = request.GET.get('default_field', '*')
+
+    sort_order = request.GET.get('sort', '').split()
+    if sort_order and len(sort_order) == 2:
+        sort = {sort_order[0]: {"order": sort_order[1]}}
+        json_query["sort"] = sort
+
+    #page?
```
michaelwood, line 587: left in by accident perhaps or the start of a comment?

grantnav/frontend/views.py
```diff
@@ -491,6 +491,185 @@ def add_advanced_search_information_in_context(context):
...
+def non_term_facet_size_from_parameters(request, json_query, agg_name):
+
+    more = request.GET.get(agg_name+'More')
+    if more:
+        json_query['extra_context'][agg_name + '_facet_size'] = 50
+    else:
+        json_query['extra_context'][agg_name + '_facet_size'] = 3
+
+
+def create_json_query_from_parameters(request):
```
michaelwood, line 577: Would be nice to have a doc string for this function as it is now heavily used. `parameters` and `json_query` are quite generic terms so for someone not familiar with the code it'd be nice to explain `json_query` =  the elastic search json query and `parameters`  = GET parameters from the request.

michaelwood (commented):
I've had a play with this it looks like a really good improvement. I didn't find any issues with it after trying various combinations of searches and orders and facets etc.

Left a few comments, will also need to be run through flake8 to pass CI. The second commit could also be squashed I think? (Initially I was confused as to why there was an unused `is_json` parameter in the functions in the first commit).

grantnav/frontend/templates/search.html
kindly: Did not want this form to have the same names as the params that end up in the URL.  Otherwise there is no way way for the backend to detect if it was this form being submitted or if the GET parameters were just comming from the URL.
After the form is submitted there is a redirect so the params in the URL turn to the correct ones.

grantnav/frontend/views.py
kindly: yep accident was not sure if we needed to do something with page here.

grantnav/frontend/tests.py
kindly: kept some more of the test

michaelwood (approved):

PR #585, opened 2020-11-17
ci: Travis -> GitHub Actions

ghost:


odscjames:
There are 2 broken runs as well.

odscjames:
[Switching from Travis to GH Actions](https://trello.com/c/oz6updTN/127-switching-from-travis-to-gh-actions)

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
+        architecture: x64
+    # This is from https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch
```
michaelwood, line 14: What about using something like https://github.com/marketplace/actions/setup-elasticsearch  ?

odscjames: Ah, didn't think to check unofficial ones. I checked the official one and that says "DONT USE THIS YET". https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch . I wouldn't rush to use an unofficial one myself.

michaelwood: Fair enough

michaelwood (approved):

Issue #594, opened 2021-02-09
Implement 360Giving Design System


robredpath:
- It's clear how the world will be different when the work is completed

All existing pages in GrantNav will use the new theme except those which are being removed. 

- It's clear what will and won't be done

Apply the [360Giving Design System](https://github.com/ThreeSixtyGiving/360-ds) to all pages on GrantNav, once [this PR](https://github.com/ThreeSixtyGiving/360-ds/pull/22) is merged

@robredpath will provide a list of pages which are being removed. 

- Any uncertainty, assumptions, dependencies or constraints are recorded

- There's an order-of-magnitude estimate of how long it will take

Days

- The work's been approved by someone at ThreeSixtyGiving

- There are links to everything that's written down about the work - such as plan.io tickets, GitHub issues

robredpath:
[...]

PR #626, opened 2021-03-24
Design updates for current sprint

michaelwood:
This is the squashed and combined version of 594/design-update--filters which at the time of branching was at commit 9a7b326bbff712043624d8cc22606a2abff31946. The 594/design-update--filters branch was in turn built on top of 594/design-update.

For issue #594


grantnav/frontend/templates/funders.html
```diff
@@ -5,35 +5,48 @@
...
+<div class="layout layout--single-column">
+  <div class="grantnav-page__content layout__content">
+
+    <div class="section">
+      <h1 class="page-title">Funders</h1>
+    </div>
+
+    <div class="section">
+      <div class="grantnav-datatable__content--filters">
+        [text-here]
```
Lathrisk, line 17: This will need to be removed, ideally before launch.

grantnav/frontend/templates/recipients.html
```diff
@@ -5,35 +5,47 @@
...
+<div class="layout layout--single-column layout">
+  <div class="grantnav-page__content layout__content">
+
+    <div class="section">
+      <h1 class="page-title">Recipients</h1>
+    </div>
+
+    <div class="section">
+      <div class="grantnav-datatable__content--filters">
+        [text-here]
```
Lathrisk, line 17: This will need to be removed, ideally before launch.

Lathrisk: N.B - no text is replacing this. It will remain empty.

grantnav/frontend/templates/funders.html
Lathrisk: N.B - no text is replacing this. It will remain empty.

grantnav/frontend/templates/funder.html
```diff
@@ -1,194 +1,228 @@
...
+  <main class="layout__content">
+    <div class="grid grid--two-columns spacer-4">
+      <div class="grid__1">
+        <h1 class="funder-header text-center">{{ funder.name }}</h1>
       </div>
-      <div class="col-xs-6">
-        {% include 'currency_stats.html' %}
+      <div class="grid grid--two-columns grid__1">
+        <div class="grid__1">
+          <dl class="top-stats">
```
Lathrisk, line 13: It would be nice if these tables looked a little nicer, but perhaps not a blocker and one to pick up on the other side (it is quite messy though). We _could_ split the `<dt>` and `<dd>` elements by line rather than wrapping them on the same line.

Lathrisk: It'll likely take up more space but I think look a little more polished... possibly. Until we get the new summary stats box in place

grantnav/frontend/templates/funder.html
```diff
@@ -1,194 +1,228 @@
...
-   <script type="text/javascript" src="https://cdn.datatables.net/1.10.10/js/dataTables.bootstrap.min.js"></script>
-   <script type="text/javascript" src="https://cdn.datatables.net/responsive/2.0.2/js/dataTables.responsive.min.js"></script>
-   <script type="text/javascript" src="https://cdn.datatables.net/scroller/1.4.1/js/dataTables.scroller.min.js"></script>
+<script
+  type="text/javascript"
+  src="https://cdn.datatables.net/1.10.10/js/jquery.dataTables.min.js"
+></script>
+<script
+  type="text/javascript"
+  src="https://cdn.datatables.net/1.10.10/js/dataTables.bootstrap.min.js"
```
Lathrisk, line 133: I think datatables depends on bootstrap, which we don't use any more so some of the icons and table functionality is broken. We might want to replace this with something else, at some point.

Lathrisk:
I've left a few comments, but more for visibility and awareness than any concrete changes (at present)

grantnav/frontend/templates/funder.html
michaelwood: Need a concrete action on resolving this if it is a blocker, otherwise we can do this as an incremental improvement after launch.

grantnav/frontend/templates/funder.html
michaelwood: I'm not seeing this as an issue on the dev site. On the datatables docs it states:

>the following Javascript library files are loaded for use in this [bootstrap-4] example:
>
>https://code.jquery.com/jquery-3.5.1.js
>https://cdn.datatables.net/1.10.24/js/jquery.dataTables.min.js
>https://cdn.datatables.net/1.10.24/js/dataTables.bootstrap4.min.js

So the only direct dependency is jquery which we do have. I we don't need the `dataTables.bootstrap4.min.js` extension as this seems to just insert classes which don't exist in our CSS, styling is currently being provided by the main theme css. 

The sort arrow / filter icons are however being provided by  `dataTables.bootstrap.min.css`

I suggest that we remove the redundant js extension and keep as is.

michaelwood: Added f162d29

BibianaC (approved):

PR #649, opened 2021-03-29
frontend: templates: base: add disableCookies to Matomo config

michaelwood:
We're not using cookie based analytics so disable this client side.


idlemoor:
that'll do nicely

idlemoor (approved):

PR #660, opened 2021-03-30
docs/terms: Update Matomo links in docs and terms iframe

michaelwood:
Also update in fail.html. Not sure if this file is active anywhere
though.

### DIFF
--- a/grantnav/frontend/templates/terms.html
+++ b/grantnav/frontend/templates/terms.html
@@ -116,7 +116,7 @@
 
         <p>{% blocktrans %}Matomo also has its own opt out mechanism:{% endblocktrans %}</p>
         <!-- opt out iframe - clicking this will mean people can opt out of tracking -->
-        <iframe style="border: 1; height: 150px; width: 600px;" src="https://mon.opendataservices.coop/piwik/index.php?module=CoreAdminHome&amp;action=optOut&amp;language=en"></iframe>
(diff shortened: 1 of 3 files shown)


grantnav/frontend/templates/terms.html
```diff
@@ -116,7 +116,7 @@ <h3>Understanding website visitor and traffic patterns</h3>
 
         <p>{% blocktrans %}Matomo also has its own opt out mechanism:{% endblocktrans %}</p>
         <!-- opt out iframe - clicking this will mean people can opt out of tracking -->
-        <iframe style="border: 1; height: 150px; width: 600px;" src="https://mon.opendataservices.coop/piwik/index.php?module=CoreAdminHome&amp;action=optOut&amp;language=en"></iframe>
+        <iframe style="border: 1; height: 150px; width: 600px;" src="https://matomo.opendataservices.coop/piwik/index.php?module=CoreAdminHome&amp;action=optOut&amp;language=en"></iframe>
```
robredpath, line 119: @michaelwood does this one need to be `/threesixtygiving/` as well?

robredpath (changes requested):
terms.html path needs updating, I think

grantnav/frontend/templates/terms.html
michaelwood: Thanks, yes it does, good spot

robredpath (approved):

Issue #683, opened 2021-04-28
Reinstate the build number / link to the footer 


stevieflow:
Via https://github.com/ThreeSixtyGiving/grantnav/issues/679

> We also used to have a link to the exact build being currently used - it might be useful to put this back into the footer.

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Working down post new design GrantNav issues](https://trello.com/c/Yz3uiR88/147-working-down-post-new-design-grantnav-issues)

odscjames:
Can this be closed now?

BibianaC:
This is now deployed.

PR #687, opened 2021-05-14
Add build number in footer

BibianaC:
Relates to #683 

Screen Shot 2021-05-14 at 15 01 11

### DIFF
--- a/grantnav/frontend/templates/components/footer.html
+++ b/grantnav/frontend/templates/components/footer.html
@@ -83,7 +83,11 @@
   <div class="footer__row wrapper footer__small-print">
 
     <div class="footer__column-2">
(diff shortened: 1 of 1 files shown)


BibianaC:
@drkane @mariongalley I have added the build number at the bottom of the footer. Is this an appropriate location or would you prefer somewhere else?

mariongalley:
Hi @BibianaC, I'm not sure if I've understood correctly, but I thought the plan might be to replace the generic Github link already in the footer that's linked to from the little icon. Would it make sense to include the link there? 
image

BibianaC:
@mariongalley what I understood, I might not be correct, is that the GitHub icon would link to the GrantNav repo in GitHub. Now it is linked to the generic 360 GitHub. And that we would reinstate the build number as it was in the previous GN website. The build number was shown like the screen shot previously.

mariongalley:
Ah yes, that makes sense. In that case at the bottom beneath the copyright warning seems reasonable to me.

BibianaC:
thanks @mariongalley

michaelwood (approved):

Issue #679, opened 2021-04-21
Footer link to Github does not link to the repo


stevieflow:
This isn't a "bug", but neither is it a feature request

The footer link for GitHub is to https://github.com/threesixtygiving/.  This then requires me to navigate to where I think grantnav code/issues might be

I can understand if the github logo/link is being used for wider context, but this doesnt seem right.  If you're linking me from grantnav to GitHub, I'd expect to get to https://github.com/ThreeSixtyGiving/grantnav/

Or - even better - I'd probably expect some text to explain that the code is on GrantNav, and then a link to https://github.com/ThreeSixtyGiving/grantnav/issues/ for me to log any feedback...

drkane:
Agreed, this is definitely worth thinking about - and there may be a better place to put the links for raising issues. 

We also used to have a link to the exact build being currently used - it might be useful to put this back into the footer.

mariongalley:
Agree it makes more sense to link specifically to the GrantNav Github rather than general 360Giving Github in the footer of GrantNav.

Re: linking the exact build being currently used, beginner question here @drkane but is that something that would need regularly updating, or would it dynamically update?

We also have copy saying to email the grantnav@ inbox with questions or feedback, which could plausibly cover the same kinds of issues one might raise in Github. We've discussed elsewhere having clearer language about when to email which inbox (e.g. support@ or info@ rather than grantnav@), so we could update the footer copy to indicate what kinds of issues we expect to see raised in Github vs via email as well?

stevieflow:
Thanks @drkane - hi @mariongalley !

OK - so this looks to be:

- link to the correct github repo - agreed
- reinstate the build number - agreed.  @mariongalley in the previous GrantNav this was autogenerated
- contact workflow for issues and feedback:  I'd advocate for both a link to the *existing* issues and detail on how to contact 360 directly.  The advantage of having access to the open issues is that people can see what has already been logged and suggested - and so may find their issue is the same/similar.

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Working down post new design GrantNav issues](https://trello.com/c/Yz3uiR88/147-working-down-post-new-design-grantnav-issues)

BibianaC:
@stevieflow What does exactly need to be done with `contact workflow for issues and feedback`.

stevieflow:
Thanks @BibianaC 

> What does exactly need to be done with contact workflow for issues and feedback

[...]

PR #689, opened 2021-05-19
Add GH issues link in footer

BibianaC:
relates to #679


BibianaC:
@mariongalley I have only added in the text `or raise an issue in Github` with a link to GH issues. Is it ok or would you prefer to also add `The existing list of issues can be found [here]`

Screen Shot 2021-05-19 at 12 42 04

mariongalley:
I guess there's fairly limited space and the GrantNav column is already the tallest. I don't mind  the change to the wording, @stevieflow would you agree?

stevieflow:
Thanks - that seems great to me @mariongalley @BibianaC

michaelwood (approved):

PR #706, opened 2021-06-01
docker: Add tests

ghost:


odscjames:
This does NOT run tests_functional - but then they don't run in GitHub Actions either??!

Lathrisk (approved):

Issue #641, opened 2021-03-25
Move divider on results card to the right


robredpath:
to make better use of the space

odscjames:
How about 70%? (Note this is with the description truncating feature of #618 & #640 too, which is still a WIP) 

![70-percent](https://user-images.githubusercontent.com/37832009/120311246-2f267380-c2cf-11eb-9953-123e285b6c42.png)

robredpath:
@odscjames 70% looks good - can you just double check that it still looks ok on smaller screens?

odscjames:
Huh - so I had just read the CSS before and thought it would be fine (it has a breakpoint in) but now I actually test small screens I find a issue. However this is an issue that exists at both 50% and 70% so I've spun off #717 to deal with it separately.

Issue #633, opened 2021-03-25
Add a link to the search bar help from the home page - above and to the right of the search bar


robredpath:

Lathrisk:
Potentially use wording from results page on the home page

BibianaC:
Link: https://help.grantnav.threesixtygiving.org/en/latest/search_bar.html

odscjames:
How is this? (narrow and wide screens)

![633-2](https://user-images.githubusercontent.com/37832009/120474129-ba6f3a00-c39f-11eb-9bd0-27e49173d247.png)


![633-1](https://user-images.githubusercontent.com/37832009/120474132-bb07d080-c39f-11eb-8a0a-e2a0c2639f01.png)

mariongalley:
Yep, looks great

PR #719, opened 2021-06-03
frontend: Newest Design System files

ghost:
https://github.com/ThreeSixtyGiving/grantnav/issues/633
https://github.com/ThreeSixtyGiving/grantnav/issues/641

These changes reflect the changes in:
https://github.com/ThreeSixtyGiving/360-ds/pull/35
https://github.com/ThreeSixtyGiving/360-ds/pull/34

It is not a straight copy of the newest scss files from the design
system over the files in GrantNav, because there are other changes
between the 2. I have avoided altering those changes as I'm not sure
which are correct.

### DIFF
--- a/grantnav/frontend/sass/grant-search-result.scss
+++ b/grantnav/frontend/sass/grant-search-result.scss
@@ -27,7 +27,7 @@
     @include breakpoint($medium) {
       padding-bottom: 0;
       margin-bottom: 0;
-      width: 50%;
+      width: 70%;
       padding-right: 24px;
       margin-right: 24px;
     }
--- a/grantnav/frontend/sass/search-block.scss
+++ b/grantnav/frontend/sass/search-block.scss
@@ -83,6 +83,18 @@
       cursor: pointer;
     }
   }
+
+  .search-block__label {
+    display: flex;
+    flex-direction: column;
+    margin-bottom: 6px;
+
+    @include breakpoint($regular) {
+      flex-direction: row-reverse;
+      justify-content: space-between;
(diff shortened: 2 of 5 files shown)


michaelwood (changes requested):
There shouldn't be any divergence between upstream 360-ds so we need to synchronise these and just pull in the changes.

When this is done could we update the date string in the base.html template (which we're doing manually at the moment (reasons)) to have the latest date to force browsers to fetch the latest changes rather than cached. Thanks 
https://github.com/ThreeSixtyGiving/grantnav/blob/master/grantnav/frontend/templates/base.html#L22

Lathrisk:
@michaelwood what does "we need to synchronise these and just pull in the changes" mean? I don't think we have a system for that at the moment. The changes described in this PR have been added to the design system in the PRs that James has noted in the comment. 

James has also created an issue around aligning the DS and Grantnav in #721 - I think we're blocked on quite a few of these issues if we need to resolve that issue first.

michaelwood:
> @michaelwood what does "we need to synchronise these and just pull in the changes" mean? 

Basically what you've said in #721, from what I understood from James' comment, this PR doesn't bring in other changes in the upstream 360-ds. I'm saying we should bring them in as a matter of course.

BibianaC:
@michaelwood @Lathrisk I have updated the theme.css date and checked the changes. Everything looks good to me so it is ready to be approved.

Note: We have decided to check the differences in styling separately.

michaelwood (approved):

Issue #685, opened 2021-05-14
GrantNav export recipients data page as CSV  giving error


michaelwood:
On https://grantnav.threesixtygiving.org/recipients there is a link to download all the data as a CSV https://grantnav.threesixtygiving.org/funder_recipients_datatables.csv

Due to the size of the data now this query to ElasticSearch exceeds the default `search.max_buckets` value and errors out. As a short term fix we can update this limit to `search.max_buckets: 200000` in the elasticsearch config. For these larger tables we may want to look at how to do offline processing to create the files and serve them statically.

michaelwood:
![](https://github.trello.services/images/mini-trello-icon.png) [GrantNav export recipients data page as CSV giving error](https://trello.com/c/h8g88gRq/160-grantnav-export-recipients-data-page-as-csv-giving-error)

odscjames:
I've done 2 commits on branches (see above) - I can work with someone next week to get these live carefully.

PR #720, opened 2021-06-04
docker: Change Elasticsearch config

ghost:
https://github.com/ThreeSixtyGiving/grantnav/issues/685


kindly:
This pull request has been superceded by #724.

This approach was preferred as it will work with any grantnav instance after a load is done and is a setting all instances that have a full load will need.

Issue #610, opened 2021-03-09
Funder page - alignment of stats on right hand side


drkane:
![image](https://user-images.githubusercontent.com/1049910/110493711-c9b95100-80ea-11eb-929e-fcdc1f88fd7f.png)

Amounts should be right-aligned. Dates in the first column look odd.

Suggest this is upgraded to something based on the new stats box when this is available.

drkane:
![](https://github.trello.services/images/mini-trello-icon.png) [GrantNav summary pages](https://trello.com/c/VMO0SMgO/155-grantnav-summary-pages)

odscjames:
> based on the new stats box when this is available.

Can I ask if this is now available, and if so where I would see it? Not sure what "new" means.

robredpath:
@odscjames it's at https://cdn.threesixtygiving.org/components/detail/search-summary.html

odscjames:
Ok so looking at https://cdn.threesixtygiving.org/components/detail/search-summary.html this is intended to replace the search box on the normal search page. I'd either have to make code changes to pull both style of search box in at once, or I'd have to change both search boxes at once. 

**Should we change both search results summaries boxes at once?** 

https://github.com/ThreeSixtyGiving/grantnav/issues/597#issuecomment-783281674 comment says we shouldn't, so I'm not clear on what needs doing.

Or instead I could look into directly fixing the individual points mentioned at the top of this thread without bringing in the new style?

ps. This is about funder pages but I think also applies to Recipient pages now? http://grantnav.threesixtygiving.org/recipient/360G-cabinetoffice-org%3AMHCLG-PH-ID-0000016

odscjames:
I posted a comment which I then deleted as I realised I was confused by something (if anyone is confused by email).

I noticed however something has to be renamed, so I did a small PR #736 to do that

odscjames:
I've started trying to apply https://cdn.threesixtygiving.org/components/detail/search-summary.html but we are going to need much more spec details before we can finish this up. 

This design component assumes there are graphs on the left side and all stats are on the right:
* For Search, Funders and Recipient pages we need defined what graphs should be on the left?
* Currently our layout of stats is over 2 columns; so do we just put all these stats in the right hand column now?
* We also need to check the interaction of the currency modal with this new look

This could take a while to sort out; **in the meantime, should I just see if I can quickly fix the issues noted at the top of this thread in the existing boxes?**

[...]

PR #736, opened 2021-06-16
frontend: templates & Sass: Rename to match Design System

ghost:
Rename to match design system component "Search summary description"
https://cdn.threesixtygiving.org/components/detail/search-summary-description.html

This is needed because we may shortly be introducing "Search summary"
https://cdn.threesixtygiving.org/components/detail/search-summary.html
so to avoid a clash with that, rename this now.

Related to https://github.com/ThreeSixtyGiving/grantnav/issues/610


Lathrisk:
Thanks, James, I've started the summary box work and done this as part of that

kindly (approved):

PR #739, opened 2021-06-17
ci: Add Compile CSS check

ghost:
This makes sure the compiled CSS that is checked in
actually matches what the compiled CSS should be.


odscjames:
To see an example of it breaking, see https://github.com/ThreeSixtyGiving/grantnav/runs/2847204563?check_suite_focus=true

kindly (approved):

PR #754, opened 2021-06-22
frontend: views: Only allow searches with allowed default fields

ghost:
We have seen cases on live server where people tried to search with odd
default fields: eg

default_field = *'A=0

In this case, the app tried to do the search then crashed when putting
together the view (in get_radio_items).

Instead, make sure we check that passed values are actually allowed.

### DIFF
--- a/grantnav/frontend/views.py
+++ b/grantnav/frontend/views.py
@@ -79,6 +79,13 @@ SEARCH_SUMMARY_AGGREGATES = {
     "max_date": {"max": {"field": "awardDate"}}
 }
 
+DEFAULT_FIELDS = [
+    {"value": "*", "name": "All grant fields"},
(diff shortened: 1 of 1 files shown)


odscjames:
It's true that this PR doesn't actually change anything from the point of the user (before they would have seen a crash, now they see a crash).

But I looked in the server logs; almost certainly this is a bot checking for SQL injection vulnerabilities so we don't care.

At least now we don't do an elasticsearch operation with bad data and the crash is recorded properly and we know the fault is with the request, not with any code we have.

BibianaC:
@odscjames @kindly what is the situation with this PR?

michaelwood:
Converting this to a draft for now as due to the conflicts/unknown state it's not ready for review at this time.

michaelwood:
Tidying up - Closing this PR as there has not been any further action for 6 months+.

PR #755, opened 2021-06-22
frontend: views: Add retry decorator on get_results

michaelwood:
This decorator allows us to handle connection issues caused by too much
traffic to the elasticsearch server.


kindly (approved):

michaelwood:
[Fix GN load breaking issue](https://trello.com/c/MorSJ3we/170-fix-gn-load-breaking-issue)

Issue #771, opened 2021-06-28
Search summary: Date awarded graph improvements


Lathrisk:
Consider ways of visualising the data before a given date. 
~~Suggested 20 years is visualised~~, with an additional bar for all grants before that date.

Lathrisk:
The recent pull request updates the graphs so that the past 20 years are shown based on today's date.
There is not currently a bar to show grants from before this date - the graph is currently laid out automatically based on the numerical order of years so the bar position and label will need to be manually entered.

Lathrisk:
@mariongalley - I have had a quick look at the years awarded graph. I'm trying to work out the best format for showing the data before our 20 year cut off. This is what I have so far. The tricky part is working out the best copy for the bar. I've currently opted for `1996+` where the range is shown in more detail in the tooltip. If I add both years to the x-axis labels then they take up too much space really.

Any suggestions?

![early_years](https://user-images.githubusercontent.com/517804/125302340-c1388780-e323-11eb-84f6-bdd551a9a504.png)

![early_years_tooltip](https://user-images.githubusercontent.com/517804/125302350-c3024b00-e323-11eb-8081-6a8cd708bb78.png)

Lathrisk:
![](https://github.trello.services/images/mini-trello-icon.png) [Summary Stats updates: Years](https://trello.com/c/J76566pW/186-summary-stats-updates-years)

mariongalley:
> I'm trying to work out the best format for showing the data before our 20 year cut off. This is what I have so far.

That's absolutely perfect, exactly what I had in mind!

> The tricky part is working out the best copy for the bar. I've currently opted for 1996+ where the range is shown in more detail in the tooltip. If I add both years to the x-axis labels then they take up too much space really.

Agree, I don't think it's necessary to add both years. I would have probably put "Pre-2003" or similar rather than "1996+" because then it's more obvious that it's all the data before the years already displayed, even if it's in some way less informative as we already know 2003 is the earliest year being displayed. Or we could say "Older data", for consistency with the Dashboard:
![image](https://user-images.githubusercontent.com/82958988/125313150-debe1f00-e32c-11eb-9b65-cc30c98eb977.png)

Lathrisk:
@mariongalley I've updated with Older data, it is a noticably longer string, but if I adjust the size of the charts it works ok:

[...]

PR #774, opened 2021-07-06
Graph tweaks

### LINKED ISSUE #770: Search summary: Amount awarded graph improvements
Use abbreviated format numbers - i.e. £1m vs £1,000,000
This will simplify the visual information and potentially allow more space for additional labels.

Lathrisk:
Closes #770 #771


requirements.txt
```diff
@@ -38,6 +38,8 @@ jsonref==0.2
     # via flattentool
 lxml==4.6.3
     # via flattentool
+millify==0.1.1
+    # via -r requirements.in
```
michaelwood, line 42: Could we just use the "humanize" django module for? -https://docs.djangoproject.com/en/3.2/ref/contrib/humanize/

michaelwood: Ah I see that it's actually doing the 10k etc

michaelwood (approved):

Issue #730, opened 2021-06-11
Static pages review copy


BibianaC:
Review copy for branding and language/tone etc.

Parent ticket #631

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Static pages review copy](https://trello.com/c/Y4DGAzp3/191-static-pages-review-copy)

PR #783, opened 2021-09-16
Funders total amount only in pounds

BibianaC:
relates to #730 

Screen Shot 2021-09-16 at 16 25 58

### DIFF
--- a/grantnav/frontend/templates/funders.html
+++ b/grantnav/frontend/templates/funders.html
@@ -12,6 +12,15 @@
       <h1 class="page-title">Funders</h1>
     </div>
 
+    <div class='spacer-4'>
+      <div class="box box--teal">
+        <h3 class="box__heading">Total Amount</h3>
+
+        <p>Total amounts shown are based on grants awarded in British pounds sterling (&pound;) only.
+          This means funders awarding grants in other currencies are not listed.</p>
+      </div>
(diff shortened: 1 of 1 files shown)


BibianaC:
@mariongalley what do you think?

BibianaC:
[Funders copy - includes £ only](https://trello.com/c/12sJfVDC/210-funders-copy-includes-only)

mariongalley:
@BibianaC I think we need to be a bit more explicit that not only are the amounts only reflecting £ grants (for those funders that do several currencies), but that funders who offer non-£ grants only are missing from the list, which is the main problem we're experiencing.

The question we get is "Why does the number of funders on this page not match the number of funders on the GrantNav homepage?"

Proposed copy from Katherine: "Total amounts shown are based on grants awarded in British pounds sterling (£) only. This means funders awarding grants in other currencies may not be listed."

BibianaC:
Thanks for the explanation @mariongalley . I have changed the copy.

mariongalley:
Sorry, bad timing, I **just** edited my comment to change the language from "may not be listed" to "are not listed". Please could you update it to say that?

BibianaC:
@mariongalley I have updated it.

michaelwood (approved):

Issue #681, opened 2021-04-22
Use 'canonical name' from data store additional data for Recipient page header


KDuerden:
**Is your feature request related to a problem? Please describe.**
GrantNav uses the registered name of a charity recipient based on Charity Commission for England & Wales data, if the recipient has a GB-CHC organisation identifier.

This feature provides a consistent and accurate name, whatever variations there might be in the published data.

However for other org-id types GrantNav uses the first version of the recipient name it comes to, which leads to issues with potentially misleading results.

Example: http://grantnav.threesixtygiving.org/recipient/GB-SC-SC001282
The page version of the name is: An Comunn Gàidhealach Meur Ìle (Islay Local Mòd Committee)
The page name is derived from this grant: http://grantnav.threesixtygiving.org/grant/360G-COOPGROUP-7522

However 'additional_data: recipientOrganizationCanonical: name' field in additional data has the official name 'An Comunn Gàidhealach'.
All the other grants published about this organisation use the official name.

This issue also appears a lot with universities, as funders often include the name of a department that received the funding:
http://grantnav.threesixtygiving.org/recipient/GB-EDU-133863
http://grantnav.threesixtygiving.org/recipient/GB-EDU-133801

**Describe the solution you'd like**
GrantNav should start to use 'additional_data: recipientOrganizationCanonical: name' field value whenever available for the recipient page name and recipient filter.

**Describe alternatives you've considered**

**Additional context**

BibianaC:
https://trello.com/c/PhwzyDac/162-use-canonical-name-from-data-store-additional-data-for-recipient-page-header

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Use 'canonical name' from data store additional data for Recipient page header](https://trello.com/c/PhwzyDac/162-use-canonical-name-from-data-store-additional-data-for-recipient-page-header)

KDuerden:
This issue has presented itself as an issue again (another Scottish charity) with an organisation requesting an update to how their name appears.

PR #786, opened 2021-09-23
dataload: import_to_elasticsearch: use canonical name.

BibianaC:
relates to https://github.com/ThreeSixtyGiving/grantnav/issues/681


Bjwebb (approved):
Looks good. Note that this requires re-importing into elasticsearch (as does my open PR https://github.com/ThreeSixtyGiving/grantnav/pull/788).

This doesn't record a duplicate if there's no `found_name`, and `name` and `org_name` don't match. I'm not sure if we want to or not. Maybe check next week with @michaelwood how the duplicates csv actually gets used?

PR #788, opened 2021-09-23
frontend: Add a Programme Title facet

Bjwebb:
Note that this requires re-importing into elasticsearch.


michaelwood (approved):
Thanks lgtm

Issue #392, opened 2017-08-04
GrantNav handles identifiers with a / in two different ways


stevieflow:
(NB - I only have an example in staging atm)

If an identifier is structured as so: ``12345/678`` then GN handles it in two ways

Via search results - the link to this grant resolves to a page that is viewable - ie: ``grant/12345/678``

Via a table results page (eg Funder page) - the link resolves to a page that is not viewable - ie: ``grant/12345%2F678``

It could be an idea to fix the latter.  

NB: 360 will be providing guidance that identifiers should *not* include ``/``

cc/ @KDuerden @morchickit

robredpath:
In the short term, we might want to fix this in the datagetter or the check_grantanv_assumption.py script that does the checks that data is GN-friendly

KDuerden:
Are there any other places where / would be problematic? I've just seen data with titles which all include / so could this break those links?
Also any other characters that might break links - are dots ok?

Bjwebb:
BTW, the link that is broken is created by a third party JavaScript library we're using for the table pages. This means that it's possibly more difficult to fix, and harder for us to say what characters may break it.

However, from experience slashes are a particular kind of problematic, so I wouldn't expect any other characters to break it. Dots should be fine.

KDuerden:
Hi, returning to slash issue. Will this Funding Org ID (taken from Open Knowledge Public bodies list) cause issues with GrantNav?  XI-PB-gb/london-councils
http://publicbodies.org/gb/london-councils

Bjwebb:
Yes, I'd expect that ID to cause a problem.

robredpath:
@michaelwood @kindly is this relevant to the current data services sprint?

robredpath:
![](https://github.trello.services/images/mini-trello-icon.png) [Data services should reject data with '/'s in](https://trello.com/c/yvNseaXs/30-data-services-should-reject-data-with-s-in)

KDuerden:
@robredpath this has just cropped up with a new publisher - their grant identifier includes /. 
We will request a change but I wanted to check what the current situation is.
Data with slash in ID or Org ID will load into data services and GN but will break in ways described in the top post?

morchickit:
HI @robredpath and @michaelwood - can you answer Katherin's question above please? :-)

KDuerden:
I've got a new publisher case where XI-PB list is only external identifier available for them. Is a Funding Org:Identifier with / still potentially problematic? 
[...]

PR #794, opened 2021-11-04
handle special characters

BibianaC:
relates to https://github.com/ThreeSixtyGiving/grantnav/issues/392


BibianaC:
[GrantNav should handle special characters in org-ids (and not URLescaping any special characters)](https://trello.com/c/yvNseaXs/30-grantnav-should-handle-special-characters-in-org-ids-and-not-urlescaping-any-special-characters)

PR #840, opened 2022-03-28
docs: add cookie consent and new analytics

robredpath:


robredpath:
So this doesn't quite look right:

Screenshot 2022-03-28 at 13 45 58

compared to how it looks in the design system:

Screenshot 2022-03-28 at 13 44 50

robredpath:
@michaelwood unless this has a super straightforward fix I'd be inclined to run with it as is, and fix later so that we can actually get the new analytics out (cc @mariongalley )

BibianaC:
The border can be fix by adding:

```
.base-card--new {
  --card-bg-hsl: var(--base-hsl);
  background: hsl(var(--orange-hsl), 0.2);
  border-left: 4px solid #DE6E26;
  border-right: 4px solid #DE6E26;
  padding-top: 32px;
  padding-bottom: 58px;
  /* I've opted to hide the before element and use borders for this instead.
  I didn't want to update the other base cards as to not break anything
  but I don't think a whole new element is needed. */ }
  .base-card--new::before {
    content: none; }
  .base-card--new .base-card__title {
    font-size: 33px;
    margin-bottom: 32px; }
  .base-card--new .base-card__text:last-of-type {
    margin-bottom: 0; }
  .base-card--new .button:hover {
    color: #153634;
    border-color: #A7531D; }

```

The `a` needs to be `color: hsl(var(--teal-dark-hsl), 1);`

docs/_templates/base.html
```diff
@@ -115,6 +115,47 @@
     {%- endif %}
   {%- endblock %}
   {%- block extrahead %} {% endblock %}
+
```
BibianaC, line 118: It would be better to add the `cookie-consent` in its own file in `_docs/_templates/_parts`

BibianaC:
@robredpath @michaelwood I have done some changes but not checked locally.

PR #844, opened 2022-04-13
cookies-consent: Avoid cookie consent box flicker

BibianaC:


mariongalley:
@michaelwood I still get the box flicker on GrantNav

PR #856, opened 2022-07-27
django: New setting for V3

ghost:


odscjames:
As comment says, didn't think it was needed at time and currently isn't. But it might be in future, so good to put in now to avoid warnings. Prompted by https://github.com/ThreeSixtyGiving/dataquality/pull/57

Lathrisk (approved):

michaelwood:
Deployed

Issue #852, opened 2022-07-11
Make GrantNav <> Insights link more seamless


mariongalley:
Two components:

- Front end: make calls to action more appealing and clearer
- Back end: Could pre-load the Insights results for your search in the background of GrantNav

mariongalley:
On making front end clearer:

- Call to action needs to be bigger/more visible - perhaps a button rather than a link
- Text needs to be clearer i.e. explain that this allows you to visualise your search as people might not know what view in Insights means

Lathrisk:
@mariongalley 

I've made a small adjustment to the insights link based on the points above. I've used the existing button styles in the deign system but we could create a new solid fill button if we want to make its presence bolder. Screenshots below:

![Screenshot from 2022-10-12 12-14-32](https://user-images.githubusercontent.com/517804/195330774-bd54e801-15ff-4c92-9f72-56c124f2b9d9.png)
![Screenshot from 2022-10-12 12-15-59](https://user-images.githubusercontent.com/517804/195330771-5e02e501-aba8-44ad-9eb7-0103676e22ad.png)
![Screenshot from 2022-10-12 12-25-49](https://user-images.githubusercontent.com/517804/195330767-5760bf5e-bf2b-4120-9ad0-07090e5b0c09.png)

mariongalley:
Hi Jared,

Thanks for this!

Feedback from me and Katherine:
- Filled button looks best, please can we go with that one?
- Can w get rid of the semicolon?
- Copy seems good to me, could potentially say "See more visualisations for this search in 360Insights" or similar, as this is above the summary.
- When the table view/card view thing goes live, we'd be keen to also add a "Visualise in 360Insights" in that area too, as feels like it's doing a similar thing to switching views/downloading results
- Would it be possible to put this improved design live ASAP and not wait for other developments around org pages/table view?

Lathrisk:
@mariongalley 
I've made these changes. They are currently built on top of the org pages and the widgets work but it shouldn't be too hard to extract those changes and apply to an earlier branch.

I'll update with progress today.

mariongalley:
Feedback from the rest of the team:
- New button styling: it's not clear that it's clickable compared to an underlined link
   - I'm not sure I agree personally, but perhaps it's because it's in a sentence which I think is resolved by the next point
- Can we put the whole copy inside the button i.e. "Visualise your search results in 360Insights" rather than just "360Insights"
[...]

PR #871, opened 2022-10-13
insights link: Upgrade insights links

Lathrisk:
#852

Screenshot: https://github.com/ThreeSixtyGiving/grantnav/issues/852#issuecomment-1277756998


michaelwood (commented):
Is there a corresponding update to the git submodule that the main css has been compiled from? 

Slightly more description in the commit message would also be good. I think that updating the theme css and the template changes are two atomic changes? so could be two commits.

Are we showing this link regardless of the current "narrow your search to 10,000 results or fewer" ?

Lathrisk:
@michaelwood I've added the submodule, updated the commit messages, and set the button to only render on <= 10,000 results

michaelwood (approved):

PR #890, opened 2022-10-28
insights: Make the button more responsive

Lathrisk:
#852 

This makes the insights link more responsive, by adding padding so that elements don't overlap and reducing the copy below a certain breakpoint.

We could flex-wrap the elements in the content-filters onto different rows as an alternative.

## Wide
Screenshot from 2022-10-28 15-42-27

## Narrow
Screenshot from 2022-10-28 15-42-14

## Possible (unrefined) wrap options as an alternative
Screenshot from 2022-10-28 15-45-06

### DIFF
--- a/grantnav/frontend/static/css/main.css
+++ b/grantnav/frontend/static/css/main.css
@@ -3,6 +3,12 @@
   align-items: center;
   display: flex; }
 
+.graph-container {
+  width: 80%;
+  height: 80%;
+  margin-left: auto;
+  margin-right: auto; }
+
 @media print {
   *,
   *:before,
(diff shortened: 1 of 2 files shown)


mariongalley (approved):
Thanks Jared, I'm happy with the reduced text and the screenshots look good to me.

codemacabre (approved):

michaelwood (commented):
The git submodule update appears to point to a non-existent 360-ds sha 

```
fatal: remote error: upload-pack: not our ref ecaa1fdeba6c0756240d19db54026fc6c6fbe52b
Errors during submodule fetch:
	360-ds
```

PR #894, opened 2022-11-08
search table: Switch sorting option

Lathrisk:
This removes sorting from the table itslef and moves all sorting options to a dropdown that aligns the card and table view.

### DIFF
--- a/grantnav/frontend/templates/components/grants_table.html
+++ b/grantnav/frontend/templates/components/grants_table.html
@@ -2,6 +2,7 @@
 {% load static %} 
     <div class="section">
       <div class="grantnav-datatable__content--filters {% if widget %}margin-top:0{% endif %}">
+        {% include 'components/dropdown-filter.html' with required=True options=dropdownFilterOptions %}
         {% if not widget %}
         <h4>Grants</h4>
           <div class="export-wrapper">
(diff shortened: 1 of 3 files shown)


grantnav/frontend/templatetags/frontend.py
```diff
@@ -195,8 +195,11 @@ def get_dataset(grant):
 
 @register.filter(name='get_current_sort')
 def get_current_sort(query):
-    for key, value in query['sort'].items():
-        return key + " " + value["order"]
+    if query:
+        for key, value in query['sort'].items():
+            return key + " " + value["order"]
```
michaelwood, line 200: I think it's preferred to use fstrings here, this helps protect against the case where and int or other type needs coercion into a string.  Something like
```python
return f"{key} {value['order']}" 
````

File: grantnav/frontend/views.py [hunk 2]
```diff
@@ -717,7 +717,14 @@ def search(request, template_name="search.html"):
         context['results'] = results
         context['json_query'] = json.dumps(json_query)
         context['query'] = json_query
-        context['view_mode'] = request.GET.get('view_mode')
+        view_mode = request.GET.get('view_mode')
+        if view_mode:
```
michaelwood, line 721: Not a blocker - could probably use the new := operator here? I do wonder if this could be simplified too? No problem if it can't be.

michaelwood (approved):
Approved - Add some thoughts but not blockers

grantnav/frontend/templatetags/frontend.py
Lathrisk: I only added the if/else block. I suspect it's flagged the rest on line changes. I suspect the existing concat pattern is consistent across a lot of grantnav but I agree the fstring approach is better. Should we opt for incremental improvements or consistency?

grantnav/frontend/views.py
Lathrisk: Update to the walrus operator.

PR #916, opened 2022-11-16
frontend: templates: grants_table: Fix missing parameters from download

### LINKED ISSUE #915: If you download a search in Table mode - it's the entire dataset
**Describe the bug**

If you download a search in Table mode - it's the entire dataset.

GN doesn't apply the filters so you get http://grantnav-org-pages.360dev1.vs.mythic-beasts.com/search.csv

When it is in classic mode the url for the download is so the specific filter is embedded: http://grantnav-org-pages.360dev1.vs.mythic-beasts.com/search.csv?fundingOrganization=GB-CHC-200051&view_mode=card

**To Reproduce**
1. Go to GrantNav search
2. Switch to Table view
3. Apply filters
4. Download your search
5. See that the download is unfiltered

**Expected behavior**
The download should be filtered to the user's search

michaelwood:
@Lathrisk would this work to fix #915 ?

### DIFF
--- a/grantnav/frontend/templates/components/grants_table.html
+++ b/grantnav/frontend/templates/components/grants_table.html
@@ -1,23 +1,19 @@
-{% load frontend %} 
-{% load static %} 
+{% load frontend %}
+{% load static %}
     <div class="section">
       <div class="grantnav-datatable__content--filters {% if widget %}margin-top:0{% endif %}">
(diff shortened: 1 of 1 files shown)


Lathrisk (approved):
LGTM, I tested it locally and it appears to produce the correct JSON

michaelwood:
Thanks :)

PR #921, opened 2022-11-18
Update organisations.md

KDuerden:
The Example text should be formatted as a hint/admonition rather than a new H2 section - [as seen in the Standard docs here](https://standard.threesixtygiving.org/en/latest/technical/identifiers/#organisation-identifier)

This PR is to correct the formatting, however this might need checking that it appears as intended. The goal is to look like the Irrelevant results hint [here](https://help.grantnav.threesixtygiving.org/en/latest/#search-results).

### DIFF
--- a/docs/organisations.md
+++ b/docs/organisations.md
@@ -17,10 +17,14 @@ In 360Giving data there are two parts to an Org ID:
 - **A list code**: a prefix that describes the list the identifier is taken from.
 - **An identifier** taken from that list.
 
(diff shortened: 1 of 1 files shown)


michaelwood (dismissed):

michaelwood:
Activated the preview here: https://help.grantnav.threesixtygiving.org/en/kduerden-formateg/

michaelwood (approved):

Issue #924, opened 2022-11-22
Prometheus monitoring metrics generation


michaelwood:
There is currently an issue with the prometheus monitoring endpoint when trying to generate the resulting stats

```
Request Method: GET
Request URL: http://127.0.0.1:8000/prometheus/metrics

Django Version: 3.2.14
Python Version: 3.8.15
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.humanize',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'bootstrap3',
 'grantnav.frontend',
 'grantnav.prometheus',
 'raven.contrib.django.raven_compat',
 'favicon')
Installed Middleware:
('raven.contrib.django.middleware.DjangoRestFrameworkCompatMiddleware',
 'raven.contrib.django.middleware.SentryMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'dealer.contrib.django.Middleware')



Traceback (most recent call last):
  File "/home/michael/dev/ods/threesixtygiving/grantnav/.ve/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/michael/dev/ods/threesixtygiving/grantnav/.ve/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/michael/dev/ods/threesixtygiving/grantnav/.ve/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/michael/dev/ods/threesixtygiving/grantnav/.ve/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/michael/dev/ods/threesixtygiving/grantnav/grantnav/prometheus/views.py", line 51, in get
    self._total_grants()
[...]

PR #925, opened 2022-11-22
grantnav: prometheus: Fix grant totals count

michaelwood:
Now that we have multiple datatypes we need to select grant data as the type of total we're looking for here.

Add test to test for this regression in the future

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/924


michaelwood:
@mariongalley we'll get one more alert and then once this is reviewed & deployed it should be fixed.

codemacabre (approved):

Issue #923, opened 2022-11-22
Sort order to be selectable in widget builder


mariongalley:
**Describe the bug**

Picking a sort order in the widget builder has no effect no the sort order in the widget itself. Widget defaults to Best Match.

**To Reproduce**

1. Go to widget builder
2. Make a selection of data and pick sort order "Latest date"
3. Embed widget on a web page
4. Selected sort order is ignored and defaults to best match

**Expected behavior**

Selected sort order should carry through to the widget

**Screenshots**

![image](https://user-images.githubusercontent.com/82958988/203349173-c98d7711-e82b-4a97-bac4-77a992ab7e77.png)

codemacabre:
I'm going to move the "Sort By" dropdown out of the preview and into the settings card. This should make it more intuitive and work as expected.

<img width="400" src="https://user-images.githubusercontent.com/49149591/204311129-3c74c0d6-eb43-4f10-9538-47f80ef1227f.png" alt="Screenshot of the widget builder with the sort by dropdown removed from the preview card and inserted into the settings card. Red annotations highlight the applicable changes." />

mariongalley:
> I'm going to move the "Sort By" dropdown out of the preview and into the settings card. This should make it more intuitive and work as expected.

Can it be in the preview as well as it needs to still be in the widget itself and people interacting with it should still be able to change the sort order?

codemacabre:
Like this?
<img src="https://user-images.githubusercontent.com/49149591/204515473-88c3d03f-e06c-4399-8e98-8a8bf63242ee.png" width="400px" alt=""/>

Issue #918, opened 2022-11-16
Table view: View-switching radiobuttons and Download buttons are off-screen on smaller widths


mariongalley:
**Describe the bug**

In classic view, you can see the radiobutton to switch views and the download buttons on smaller screen widths.
In card view, you have to use the scrollbar to scroll horizontally to see them.

**To Reproduce**
1. Go to GrantNav search
2. Make your screen narrower
3. Switch to card view
4. Radiobuttons and download buttons are no longer visible

**Expected behavior**

- [x]  I'd like to be able to see the buttons on any screen size
- [x]  I'd also like the Insights button to appear in Table view

**Screenshots**

Table view

![image](https://user-images.githubusercontent.com/82958988/202224476-5edf5b12-813e-41f9-94aa-4d7c362de6a5.png)

Classic view

![image](https://user-images.githubusercontent.com/82958988/202224706-d229a75c-717b-41ea-bd94-852c9914b67a.png)

codemacabre:
> **Expected behavior**
> 
> * [ ]   I'd like to be able to see the buttons on any screen size
> 
> * [ ]   I'd also like the Insights button to appear in Table view

Fixing the table responsivity (i.e. the top item) involves changes to the design system; adding an Insights button involves changes to grantnav's `grants_table.html`, so this will require a couple of PRs to complete.

PR #932, opened 2022-11-28
widgets: add sort dropdown to settings

codemacabre:
## Summary
+ Add sort dropdown to widget builder settings.
+ Add "Visualise in 360Insights" button to table view

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/923 & part of https://github.com/ThreeSixtyGiving/grantnav/issues/918

## Verify
1. Visit the widget builder
2. Make a search query and select a sort order using the "Sort by" dropdown
3. The page should reload with the selected sort applied, and reflected in the preview
4. Copy code and embed elsewhere to verify the sort order is preserved

and

1. Return to home and perform a search, using filters to limit results to < 10,000
2. "Visualise in 360Insights" button should be visible in both classic and table view

### DIFF
--- a/grantnav/frontend/static/css/main.css
+++ b/grantnav/frontend/static/css/main.css
@@ -5444,6 +5444,7 @@ a.base-card:hover:before {
       display: none; } }
 
 .grantnav-search__content {
+  overflow-x: auto;
   padding-top: 0px;
   width: 100%; }
   @media (min-width: 60em) {
--- a/grantnav/frontend/templates/components/dropdown-filter.html
+++ b/grantnav/frontend/templates/components/dropdown-filter.html
@@ -1,7 +1,11 @@
 {% load get_current_sort from frontend %}
 <form id="sort-form" class="form-inline">
   <div class="dropdown-filter">
+    {% if title %}
+    <label class="dropdown-filter__label" for="results-sort-dropdown">{{ title }}</label>
+    {% else %}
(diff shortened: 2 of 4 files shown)


michaelwood:
I had a quick test. This looks really good @codemacabre 

@mariongalley I suggested we remove the order-by in the widget itself as this could be quite confusing / undesirable if the user has deliberately chosen an order-by setting.

image

michaelwood:
Just seen your comment on #923 . If you'd definitely still like it on the widget itself, maybe we call it "Default order by" in the settings pane.

mariongalley:
> Just seen your comment on #923 . If you'd definitely still like it on the widget itself, maybe we call it "Default order by" in the settings pane.

That sounds good - yeah I think one of the things the publishers were most keen on was for the widgets to be dynamic, and this removes one of the only things the user interacting with the widget can change which seems a shame

codemacabre:
I've returned the dropdown to the widget, and made the dropdown's title customisable so we can change it on the widgets page. While I was there, I've also added the 360Insights button as requested in https://github.com/ThreeSixtyGiving/grantnav/issues/918, so this should complete both issues.

michaelwood (changes requested):
Great - Just one thing missing is that the git submodule for 360-ds is now pointing at an old commit and needs to be at the new HEAD of the 360-ds repository

michaelwood (approved):

PR #933, opened 2022-11-29
update_theme: Add helper script for updating theme files from 360ds

michaelwood:


codemacabre (dismissed):

michaelwood:
@codemacabre Is there a better way to do the `npm install` task, if the node modules are already installed it's probably good to just not update modules/update package-lock.

codemacabre:
@michaelwood There's `npm update` which will update according to the package.lock and install any missing modules: https://docs.npmjs.com/cli/v6/commands/npm-update

Bjwebb (approved):

Issue #873, opened 2022-10-14
Widgets - minor changes


mariongalley:
- [x] Would you be able to make the widget width relative (defaulting to 100%) rather than a fixed width so that it's responsive?
    - [ ] The option to choose what % would be nice, but is optional as users can change that themselves quite easily within the code
- [x] Could you remove the border around the widget?
- [x] Could we have a white background where it's currently transparent?
- [x] Could you add a 360Giving logo instead of the "Export table data" controls, and make the header less tall generally?
- [x] Could the "Grants" title be a bit more explicit e.g. "Grants data from 360Giving"? (copy TBC)
- [x] Can you put the "To explore this data in more detail, visit Grantnav" call to action at the top rather than the bottom, and potentially style it more like the new "open in 360Insights" call to action rather than as a yellow info box?

Lathrisk:
@mariongalley 
There were a few things going on in the suggestions above so I have mixed them into something cohesive, but more than happy to rework this if it has diverge from your vision of it.

These changes remove the grants table title, the info bar at the bottom, and the CSV/JSON download buttons.
These are replaced with a header that contains the title, and a link to grantnav which includes the logo.

![Screenshot from 2022-10-26 15-20-49](https://user-images.githubusercontent.com/517804/198051946-88957809-e5b2-4dcc-941d-1e1dcd3be6c7.png)

mariongalley:
> There were a few things going on in the suggestions above so I have mixed them into something cohesive, but more than happy to rework this if it has diverge from your vision of it.

@Lathrisk No I think this looks great, thank you! I've ticked off the things that are done, but I couldn't ascertain whether the width and the background colour ones were done based off this screenshot - you can tick them off if they're done.

> This header is styled based on the summary header insights link on the main search page.
>
> EDIT: Would you prefer the CTA style rather than the logo link and large header?

I really like having a nice header that clearly states that the data is from GrantNav and how to go there. However, I don't think we can put the logo over a yellow background as the contrast with the yellow dots is insufficient (and there's something in our brand guidelines about it, probably!)

I don't in principle object to the colours from the search summary being used if we can work out a way to give the logo a white background that looks nice. Otherwise we can style it more like other headers we have e.g. use a white background with text in the 360Giving colours and a light grey divider - but less tall!
![image](https://user-images.githubusercontent.com/82958988/198072836-d9e2ef14-ea9f-401e-9a01-6dbfa7023de1.png)

[...]

PR #942, opened 2022-12-13
widgets: add width controls & ESLint

codemacabre:
## Summary
+ Add form controls on widget builder to customize widget width

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/873

Also,
+ Add ESLint to statically analyse JS and `<script>` tags within HTML
+ Perform lint operation and fixes (inc. whitespace / indentation, changing syntax to ES6, etc.)

## Verify
1. Visit Widget Builder
2. In widget settings section, ensure new 'Widget width' form is displayed, including:
   + Numerical `<input>`
   + Dropdown `<select>` with 'pixels' and 'percent' options
3. Ensure making changes to Widget width controls are reflected in the `code` and preview

Optional:
+ (ESLint) Check everything JS-related still works!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,24 @@
+{
+  "root": true,
+  "env": {
+    "browser": true,
+    "es2021": true,
+    "jquery": true
+  },
+  "plugins": [
+    "html"
+  ],
+  "parserOptions": {
+    "ecmaVersion": "latest"
+  },
+  "extends": [
+    "eslint-config-semistandard"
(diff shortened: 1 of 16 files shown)


michaelwood (changes requested):
Looks fine afaics, there are a couple of Git best practice things needed here though; we want to make sure the linting actions are in their own commit and we need the branch rebased on the 'master' branch rather than updated with a merge.

Issue #959, opened 2023-01-27
Ability to limit the grants table to a small number of rows instead of showing the full results of the query


mariongalley:
**Is your feature request related to a problem? Please describe.**
Publishers want to show e.g. the 10 most recent grants, and no more, in the widgets table, to avoid pagination and scrolling.

**Describe the solution you'd like**
In the widgets builder, ability to limit the amount of rows shown for the table view.

**Describe alternatives you've considered**
Users can currently show 25, 50, 75 or a hundred rows.

codemacabre:
@mariongalley Would you like the existing ability to show 25, 50, 75 or 100 rows to be disabled / hidden when reducing the widget to `x` number of rows?

mariongalley:
@codemacabre Good question - yes I think so. The publisher in question wanted to limit the rows to stop people having an overwhelming amount of data, so yes I think the users shouldn't have the option to add more

PR #970, opened 2023-02-13
frontend: widgets: Allow number of rows to be specified

codemacabre:
## Summary
+ Add input for restricting results to specific number of rows
+ Show / hide pagination controls based on number of rows

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/959

## Verify
+ Visit the widgets builder
+ Leave the rows input empty or at 0
+ Verify the page length options still work and number of rows changes accordingly
+ Verify the pagination controls are displayed
+ Change the rows input to > 0
+ Verify the table shows the selected number of rows
+ Verify the pagination controls are hidden

### DIFF
--- a/grantnav/frontend/templates/search_widgets.html
+++ b/grantnav/frontend/templates/search_widgets.html
@@ -118,7 +118,7 @@
                 <div class="dropdown-filter flex-centre">
                   <label class="dropdown-filter__label" for="input-rows">Number of rows:</label>
                     <div class="filter-list__contents--form-item">
-                    <input class="form-control input-xs" type="number" name="rows" id="input-rows" min="0" value="" placeholder="Show all results" onchange="updateFrames()" />
+                    <input class="form-control input-xs" type="number" name="rows" id="input-rows" min="0" value="" placeholder="Show all results" onchange="updateRows()" />
                   </div>
                   <div class="alert-tag margin-left:1">
                     <span class="alert-tag__icon"><svg width="16" height="16" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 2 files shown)


grantnav/frontend/templates/search_widgets.html
```diff
@@ -211,18 +224,20 @@ <h1>Search input is not valid</h1>
...
+    const rows = inputRowsElement.value;
 
     const preview = document.getElementById('sample-frame');
     const code = document.getElementById('generated-code');
 
-    code.innerHTML = `&lt;iframe src="${window.location.origin}/widget-render/${selectedView}${window.location.search}" style="height: 700px; width: ${width}${widthType};" frameBorder="0"&gt;&lt;/iframe&gt;`;
+    code.innerHTML = `&lt;iframe src="${window.location.origin}/widget-render/${selectedView}${window.location.search}?rows=${rows}" style="height: 700px; width: ${width}${widthType};" frameBorder="0"&gt;&lt;/iframe&gt;`;
 
-    preview.src = `/widget-render/${selectedView}${window.location.search}`;
+    preview.src = `/widget-render/${selectedView}${window.location.search}?rows=${rows}`;
```
michaelwood, line 240: Not sure this looks right, a resulting search string would be something like:

`'?query=wef&default_field=%2A&sort=_score+desc?rows=3'` 

Whereas what I think would be desired is

`'?query=wef&default_field=%2A&sort=_score+desc&rows=3'`

codemacabre: Will `window.location.search` always have some query params? If so, I can just change the `?` to a `&`. If not I'll put some conditional in there.

michaelwood: Unfortunately that can't be relied on. You could do an `if` conditional but might be better to use [URLSearchParams]( https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams )

michaelwood:
> There is a slight issue here where if you change a search parameter (on the left hand side) the rows value gets reset.

Had a discussion with @codemacabre about this and this is more a UX issue with the filters/search and the display settings getting a bit visually muddled. When you change search/filters the number of rows will depend on the filter but as the page has little ordering that can be inferred this constraint isn't obvious. 

I'll raise a separate issue to track this generally.

michaelwood (approved):

michaelwood:
New issue for design work #974

Issue #957, opened 2023-01-25
Add recipient individuals count to home page headline stats


michaelwood:
Add the count of recipient individuals to the headline stats. 

Some design thought is needed here on the home page. Space is quite limited, we will want to avoid the wrapping like, or if it does wrap make sure we can make them all align the same:

![image](https://user-images.githubusercontent.com/493697/214682330-6d8388cf-b3a7-4117-ab31-9f71114f1a94.png)

Removing the word recipient:

![image](https://user-images.githubusercontent.com/493697/214682787-1aec797d-8bae-4809-ac15-154058db5a3b.png)

Slightly left field suggestion - We could also make it into a sentence:

![image](https://user-images.githubusercontent.com/493697/214684649-09b68341-ccad-44d2-8e86-221e6b1d9221.png)

This is likely going to be an issue across all the tools, "organisations" and "individuals" being rather long words.

michaelwood:
*across  / with

michaelwood:
I think multiple rows might be good option. We could neaten that up by making it a square, the grid system in the 360-design system is a little overly space-y so we could also compress that with a custom style too.

![image](https://user-images.githubusercontent.com/493697/215120453-1d549bb9-60d5-4653-8167-558dfcf867cb.png)

Or using the existing headline stats widget that we use elsewhere could also work:

![image](https://user-images.githubusercontent.com/493697/215126250-79ae2c74-ce64-4691-a457-2f67555c1132.png)


Another option is also scaling the whole headline stats to be smaller so that it won't wrap - smaller font size and so on. They start to loose their headline ability though.

michaelwood:
Also if someone can come up with a longer word than "all" in "Search All grants" then that helps with the symmetry :)

KDuerden:
Very good, these look nice.

I also like the the first option best, and I agree on the rearrangement of the links to have Funders & Grants first, with the two recipient types below.

michaelwood:
Just to double check (forgot to number the options!) first option as in (and not one in a previous message):

> I think multiple rows might be good option. We could neaten that up by making it a square, the grid system in the 360-design system is a little overly space-y so we could also compress that with a custom style too.
> 
> ![image](https://user-images.githubusercontent.com/493697/215120453-1d549bb9-60d5-4653-8167-558dfcf867cb.png)
>

mariongalley:
[...]

PR #975, opened 2023-02-20
Mw/std updates 1

michaelwood:
* Headline stats
* Search summary box
* Filter toggle for Recipient Type
* Remove link for organisation page if recipient type

(Test version available on the grantnav dev server http://grantnav.360dev1.vs.mythic-beasts.com/ 
 Note any functionality regressions (e.g. missing geo filters etc) are due to having curated tested data.)

### DIFF
--- a/grantnav/frontend/views.py
+++ b/grantnav/frontend/views.py
@@ -36,10 +36,11 @@ BASIC_FILTER = [
     {"bool": {"should": []}},  # additional_data.recipientRegionName
     {"bool": {"should": []}},  # additional_data.recipientDistrictName
     {"bool": {"should": []}},  # currency
-    {"bool": {"should": []}},   # additional_data.TSGFundingOrgType
+    {"bool": {"should": []}},  # additional_data.TSGFundingOrgType
     {"bool": {"should": {"range": {"awardDate": {}}}, "must": {}, "minimum_should_match": 1}},   # Date range
     {"bool": {"should": []}},  # Programme Title
     {"bool": {"should": []}},  # additional_data.TSGRecipientType
+    {"bool": {"should": []}},  # simple_grant_type
 ]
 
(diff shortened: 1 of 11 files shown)


michaelwood:
related issue #957

mariongalley:
Two comments from Katherine @michaelwood :
- Please can Recipient type be Expanded by default, like Funding Org Type is, as this takes some of the guessing out of what kind of type it is
- We need to call it something other than “Recipient type” as this already means something in Insights, and there's a reasonable chance that we will add Recipient Type filtering to GrantNav in the future. Suggest we call it Type of Recipient and rename the existing Recipient type in Insights to Recipient Organisation Type

michaelwood:
@mariongalley Think I've addressed those two points (dev site updated).

mariongalley:
Thanks @michaelwood - I'll check if there are any objections before we amend the name in live insights

grantnav/frontend/views.py
```diff
@@ -601,12 +615,13 @@ def search(request, template_name="search.html"):
             json_query_param = json_query_param.replace('"recipientDistrictName"', '"additional_data.recipientDistrictName"')
             json_query = json.loads(json_query_param)
             filter_ = json_query['query']['bool']['filter']
-            # There were originally 8 filters ES now expects all 11 so append
-            # the missing ones
+            # There were originally 8 filters with the old urls ES now expects all 11 so append
```
Bjwebb, line 618: Should this comment say 12 now?

michaelwood: yep thanks will change

Bjwebb (dismissed):

Bjwebb (approved):

michaelwood:
(deploy scheduled for 06/03/23)

PR #990, opened 2023-05-18
templates: grant: Add feature to show details of grant to individual

michaelwood:
Add a grant details page for individuals based on existing design:

image

The link when the grant is to a recipient organisation normally goes to the org page. The best idea I had for where the link could go to was link it to a field search `recipientIndividual.id:360G-id-2344-example`. Real example https://grantnav.threesixtygiving.org/search?query=recipientIndividual.id%3A360G-BarnwoodTrust-IND-0001_2022-06&default_field=%2A&sort=_score+desc&recipientTSGType=Individual

### DIFF
--- a/grantnav/frontend/templates/grant.html
+++ b/grantnav/frontend/templates/grant.html
@@ -22,7 +22,7 @@
       </h1>
     </div>
 
-    {% if grant.source.fundingOrganization.0.id and grant.source.recipientOrganization.0.id %}
+    {% if grant.source.fundingOrganization.0.id %}
     <div class="spacer-4">
       <div class="grant-info-highlight grid grid--three-columns">
         <div class="grant-info-highlight__funder">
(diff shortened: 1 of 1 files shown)


mariongalley:
Thanks @michaelwood, that makes sense to me. I don't think it will be very common, but some funders of individuals will re-use IDs where they give several grants to the same individual, especially for recurring grants.

michaelwood:
> Thanks @michaelwood, that makes sense to me. I don't think it will be very common, but some funders of individuals will re-use IDs where they give several grants to the same individual, especially for recurring grants.

Yeah understood, sounds like something to revisit based on how things pan out with publishers. If this good enough for now, are you able to approve it? Thanks

mariongalley (approved):
This looks good and makes sense to me

mariongalley:
>  So this means there will be no 'individual recipient' page, only grant records or cards/table view in the search?

I think that seems reasonable given there's no aggregate data about an individual we can show except for the shared identifier, so generating a page for each individual identifier wouldn't add value.

> And the individual recipient in the card view will continue to be text but not hyperlinked?

Good point, I reckon both places should behave in the same way, so either neither of them links anywhere or they both link to the same place.

PR #1005, opened 2023-09-04
settings: Namespace the static directory

michaelwood:
This avoids any potential namespace clashes on servers which have other applications running on them as this can be an URL setup in the global webserver name space.


michaelwood:
Note there is a corresponding deploy config PR for this that needs to be highstated simultaneously

R2ZER0 (approved):
Looks good :+1:

PR #1006, opened 2023-09-12
Mw/insights api

michaelwood:
Add a new API for aggregate information to power Insights


dataload/import_to_elasticsearch.py
```diff
@@ -243,7 +249,29 @@ def maybe_create_index(index_name=ES_INDEX):
                             }
                         }
                     },
-
+                    "GNCanonicalRecipientOrgId": {
```
R2ZER0, line 252: Is this elasticsearch document structure defined by any other schema, like with the CSV downloads, or is this purely internal to grantnav, and it can be whatever we choose in the same vein Django model?

dataload/import_to_elasticsearch.py
```diff
@@ -424,6 +457,50 @@ def grant_generator():
...
+    if value is None:
+        return None
+    previous_value = bins.pop(0)
+    for i, b in enumerate(bins):
+        if value > previous_value and value <= b:
+            return labels[i]
+        previous_value = b
+
+
+def update_doc_with_first_recipient_org_info(grant):
```
R2ZER0, line 488: Why do this? 
Do we ever get grants with more than one recipient org, or is it not worth thinking about?

dataload/import_to_elasticsearch.py
```diff
@@ -14,12 +14,18 @@
...
+
+
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "grantnav.settings")
 
+
 sys.path.append(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
 
 from grantnav.frontend.org_utils import new_ordered_names, new_org_ids, get_org, OrgNotFoundError # noqai
 
+django.setup()
```
R2ZER0, line 28: Can `django.setup()` safely go above the `from grantnav.frontend ... import ...` line, or does it need some env setup from that import?

I say because I once spent ages debugging the order of `django.setup()` and imports, where some imports wouldn't work properly until after setup, and since then as a general rule I've tried to put it as high up the file as possible

michaelwood: Yep happy to do that

dataload/import_to_elasticsearch.py
michaelwood: Yeah this is purely internal to grantnav, we choose what data it holds. The main difference from a django is that there are many more fields which aren't mentioned here which are created automatically/inferred from the data.

dataload/import_to_elasticsearch.py
michaelwood: Elasticsearch doesn't really like indexing on nested data (not sure it's actually possible) and we want a field in the first org info which is the best matching one (from FindThatCharity), it is possible for multiples, such as when an organisation has both a charity and a company. We extract that field to a higher level in the document object so that we can do an aggregate on the field more easily.

grantnav/frontend/views.py
```diff
@@ -626,11 +628,15 @@ def search(request, template_name="search.html"):
...
 
+    try_cache = False
+
     query = request.GET.urlencode()
     if query:
         context['query_string'] = query
     else:
         context['query_string'] = ""
+        if not settings.DEBUG:
+            try_cache = False
```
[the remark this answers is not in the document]
michaelwood: Ah nice catch, yep I left that in by accident.  Now set to true.

R2ZER0 (approved):
Looks good to me :+1:

PR #1023, opened 2023-12-20
views: filter_search_ajax Add case matching

michaelwood:
Adds new string case matching to add matching against FindThatCharity given names in the data.

Related: https://github.com/ThreeSixtyGiving/grantnav/issues/965

### DIFF
--- a/grantnav/frontend/views.py
+++ b/grantnav/frontend/views.py
@@ -884,7 +884,7 @@ def filter_search_ajax(request, parent_field=None, child_field=None):
         # for each word in the split add wildcard (*) before and after
         # allow both the original search term and a capitalized version of the word using OR
         # AND all the individual word queries together.
-        and_terms = " AND ".join([f"(*{part}* OR *{part.capitalize()}*)" for part in filter_search.split()])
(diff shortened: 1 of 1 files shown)


michaelwood:
Will need to be rebased on https://github.com/ThreeSixtyGiving/grantnav/pull/1022

michaelwood:
Had one more idea on this, so just testing that

michaelwood:
> Had one more idea on this, so just testing that

Didn't work https://github.com/ThreeSixtyGiving/grantnav/issues/965#issuecomment-1879016170

This not changing this PR, is ready for review now

R2ZER0 (approved):
Is there not an actual case-insensitive filter?

I guess not, so Looks good :+1:

michaelwood:
> Is there not an actual case-insensitive filter?
> 
> I guess not, so Looks good 👍

Yeah it's quite a bit more involved on the keyword field type, I think that's why it's been done like this in the existing bit. If we come to change this bit to support multiple alternative name matching I think this will be refactored out anyway and this way its more explicit what the matching rule is.

PR #1024, opened 2024-01-03
Update Filter reference table and fix internal references

mrshll1001:
This updates the Refining Results page of the helpsite docs with new information as well as refactors some internal links to :doc: and :ref: in appropriate places, to ensure internal linking works in RST.

The resulting page can be viewed [here](https://help.grantnav.threesixtygiving.org/en/mm-helpsite-fixes/refining_results/#).

### DIFF
--- a/docs/refining_results.md
+++ b/docs/refining_results.md
@@ -43,8 +43,8 @@ The numbers in brackets next to the values show the number of grants in the sear
     - Recipient Org:Name
     - See also: :doc:`./organisations`
   * - Funding Organisation Type
-    - tbc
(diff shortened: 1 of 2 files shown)


michaelwood (commented):
Content looks good. 

Just an FYI for future - In the dev software projects we use a scheme for commit messages to reference the logical module/area we're working in: https://github.com/ThreeSixtyGiving/.github/blob/main/developer.md#log-message , so the subject line of these commit messages would ideally be prefixed with "docs: <short message>". I'll quickly update that now.

michaelwood (approved):

PR #1025, opened 2024-01-05
Mw/insights url config

michaelwood:


michaelwood:
This is the first change to make the grantnav to insights button configurable. Further work In #1022 where the failing tests are fixed and another commit which is changing the copy for the button and removing the 10k limit however that work is not signed off for "live" currently.

grantnav/frontend/context_processors.py
```diff
@@ -29,3 +29,7 @@ def main_css_cache_key(request):
 
 def debug_mode(request):
     return {"debug": settings.DEBUG}
+
+
+def insights_url(request):
```
R2ZER0, line 34: Do we need to add this processor to `TEMPLATES[0]["OPTIONS"]["context_processors"]` in `settings.py`? Or is this used some other way? I'm not really familiar with templates

PR #1026, opened 2024-01-09
Mw/grantnav jan 2024

michaelwood:
All live-able commits from the various branches that I had


grantnav/frontend/tests_functional.py
```diff
@@ -28,9 +28,9 @@ def browser(request):
         chrome_options.add_argument("--headless")
         # no-sandbox prevents an error when running as the root user
         chrome_options.add_argument("--no-sandbox")
-        # uncomment this if "DevToolsActivePort" error
-        # chrome_options.add_argument("--remote-debugging-port=9222")
-
+        # uncomment this if "DevToolsActivePort" error / ubuntu snap workaround
+        chrome_options.add_argument("--remote-debugging-port=9222")
```
R2ZER0, line 32: Re-comment this workaround?

I copy-n-paste a whole bunch of code to make it use Firefox every time I test, there must be a better way...

grantnav/api/aggregates.py
```diff
@@ -11,7 +11,13 @@ def get(self, *args, **kwargs):
         # Append .aggregates_api as this is the flag we currently switch on in the
         # main search functions to know what format we will return results as.
         self.request.path = self.request.path + ".aggregates_api"
-        context = search(self.request)
+        ret = search(self.request)
```
R2ZER0, line 14: A view calling another view :smile: Is this a consequence of the backends merge? that `search` looks real hairy, doubt we'll be refactoring that anytime soon.

grantnav/frontend/views.py
```diff
@@ -622,6 +632,50 @@ def widget_search(request, json_query):
...
+    """
+    do_redirect = False
+    request_get_copy = request.GET.copy()
+
+    for entity_type in [("fundingOrganization", "funder"), ("recipientOrganization", "recipient")]:
+        # Match the supplied org-id to any other org-ids in use
+        if org_ids := request.GET.getlist(entity_type[0]):
+            for org_id in org_ids:
+                try:
+                    for additional_org_id in get_org(org_id, entity_type[1])["orgIDs"]:
```
R2ZER0, line 668: Does this include the canonical org id (as well as "additional" org ids), in case they've searched for a non-canonical org? (if that's even possible here?)

R2ZER0 (approved):
Looks good to me!

Just a couple comments for my understanding. Maybe recomment Chrome workaround.

grantnav/frontend/views.py
michaelwood: yeah the "orgIDs" field contains all the org-ids known for that org

grantnav/frontend/tests_functional.py
michaelwood: Argh yeah, I'm tempted to leave that in since it works in CI and I don't think it does any harm? Maybe a better way is to have our own .selenium.conf or something, it's a bit of a nightmare though as they're liable to change.

grantnav/api/aggregates.py
michaelwood: Yeah, it's overdue some rationalisation, it's not quite as scary as it looks and this can be thought of a wrapper around the search request, with a minimal amount of duct tape to avoid making the search function larger and to start the process of splitting it up.

PR #1029, opened 2024-01-10
Mw/fix helpsite links

michaelwood:
1. Rewrite the link checker test which missed the the help site broken links
2. Fix the read the docs links
3. Fixes all the links the new link checker found as broken


michaelwood:
@mariongalley after rewriting the automated link checker code and fixing the readthedocs links I found and fixed the rest of the newly found broken links, could you have a scan down the changes to check to see if there are any that you don't agree with. 

http://www.threesixtygiving.org/standard/ is now the non-redirected version to https://www.threesixtygiving.org/data-standard/ and other broken links such as ordnancesurvey and charitycomission I've searched the web for the new equivalent pages and put the logical one in.

mariongalley:
@michaelwood I'm not at all confident I understand what I'm looking at, but some of this doesn't look right to me e.g. Download URL is replaced with #

image

michaelwood:
> @michaelwood I'm not at all confident I understand what I'm looking at, but some of this doesn't look right to me e.g. Download URL is replaced with #
> 

Thanks yeah, that particular one is pretend/test data to populate the https://grantnav.threesixtygiving.org/datasets/ page.

I've replaced the links with "#" because the publishers probably don't want us causing lots of hits on their site with our automated testing if we can help it. The links are only there to add rows to the "360Giving datasets used" table for automated testing purposes so I decided it isn't worth maintaining real links.

mariongalley:
@michaelwood Right, thanks! I still didn't really get what I was looking at with the diff checker. Would it be possible for Matt or another colleague to check it?

michaelwood:
> @michaelwood Right, thanks! I still didn't really get what I was looking at with the diff checker. Would it be possible for Matt or another colleague to check it?

Thanks yep, over to @R2ZER0 for a final pass

grantnav/frontend/templates/components/search-summary-box.html
```diff
@@ -141,7 +141,7 @@
             <span class="summary-footer--item__value">
               {{currency|currency_symbol}}{{results.aggregations.earliest_grant.hits.hits.0|get:"_source"|get:"amountAwarded"|human_format}}
               from
-              <a href="grant/{{results.aggregations.earliest_grant.hits.hits.0|get:"_source"|get:"fundingOrganization"|first|get:"id"}}">
+              <a href="grant/{{results.aggregations.earliest_grant.hits.hits.0|get:"_source"|get:"id"}}">
```
R2ZER0, line 144: Can we use django `reverse` in templates?

R2ZER0 (approved):
Looks good to me :+1: 

I suppose we could also be more permissive to only check anchors with href, to allow for anchors without href, if we need to in future.

Just remember to remove the TEMP FIX after deploy :)

grantnav/frontend/templates/components/search-summary-box.html
michaelwood: That would be better yeah. Gonna leave it as out of scope for this time though.

michaelwood:
> 
> Just remember to remove the TEMP FIX after deploy :)

https://github.com/ThreeSixtyGiving/grantnav/pull/1031

PR #1031, opened 2024-01-16
tests_functional: Remove temp workaround for helpsite error

michaelwood:
Remove temp work around for a fix to help.grantnav site that is now live.


R2ZER0 (approved):
Looks good :tada:

PR #1034, opened 2024-01-25
templates: Remove 10k limit for viewing results in Insights

michaelwood:


michaelwood:
@mariongalley just FYI there will be a short period where the 10k limit is still in place as I need to only deploy that limit removal when the new insights has switched over

R2ZER0 (approved):
Looks good :+1:

PR #1066, opened 2024-05-16
Update terms.html

mariongalley:
Update now "Where is this data from" has changed to be about grantee amendments to tell people how to find the publisher and update the laws to say England & Wales only as Scotland & NI may have different applicable laws


michaelwood (approved):

michaelwood:
Thanks, now deployed

Issue #1056, opened 2024-04-08
Widgets - improve accessibility


mariongalley:
A few accessibility issues with Widgets have been pointed out by a publisher, please can these be fixed:

- [x] 1. The TITLE tag gives all widget instances the page title "360Giving Widgets". Many screen reader set ups will use the title to tell the user about the contents of the frame so it helpful when this is descriptive. Is there a way to programmatically title the page with human readable content based on the data selected for display in the widget (e.g. "Grants data from Very Nice Foundation")?

- [x] 2. There is no language defined for the widget content. This is important to screenreaders (for pronunciation).

- [x] 3. The (extremely elegant) logo implementation creates no accessible title and therefore provides no navigation scent to screenreaders. Adding aria-label="some useful text here" to the A tag around the logo will solve this. (I am unclear whether you intend the logo to link the GrantNav homepage or to the relevant dataset for the trust, as it does now).

michaelwood:
On the widget itself 

open in [GrantNav](https://grantnav.threesixtygiving.org/search)

Needs to be:

[open in GrantNav](https://grantnav.threesixtygiving.org/search)

PR #1078, opened 2024-05-29
Mw/widgets accessibility

michaelwood:
https://github.com/ThreeSixtyGiving/grantnav/issues/1056

### DIFF
--- a/grantnav/frontend/templates/base.html
+++ b/grantnav/frontend/templates/base.html
@@ -1,10 +1,10 @@
 {% load static %}
 {% load favicon %}
-<!doctype html>
+<!doctype html lang="en">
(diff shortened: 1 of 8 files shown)


codemacabre (dismissed):
Not part of your changes, but the `href` to `about:blank` seems like it would cause problems. Presumably you're handling this and rewriting the link elsewhere? I'm not sure what a screen reader would make of this if it encountered a link to a blank page.

michaelwood:
> Not part of your changes, but the `href` to `about:blank` seems like it would cause problems. Presumably you're handling this and rewriting the link elsewhere? I'm not sure what a screen reader would make of this if it encountered a link to a blank page.

Yeah that's a bit of an oddity apparently it gets set in JS https://github.com/ThreeSixtyGiving/grantnav/blob/live/grantnav/frontend/templates/widgets/grantnav_link_script.html

codemacabre (approved):
lgtm!

PR #1080, opened 2024-07-03
Mw/geo location improvements

michaelwood:


R2ZER0 (dismissed):
Looks good :+1: Approved pending updated test data, once the tests pass.

dataload/import_to_elasticsearch.py
```diff
@@ -483,6 +520,106 @@ def to_band(value, bins, labels):
         previous_value = b
 
 
+def update_doc_with_other_locations(grant):
+    """ This flattens out some embedded data for easier indexing """
+
+    # Prior versions of additional_data may not have this field
+    # or if locationLookup failed entirely for this grant
+    if not grant["additional_data"].get("locationLookup"):
```
rhiaro, line 528: is `grant["additional_data"]` always guaranteed to be set and the correct type at this point, or does this need a try/except?

dataload/import_to_elasticsearch.py
```diff
@@ -483,6 +520,106 @@ def to_band(value, bins, labels):
...
+                    grant["additional_data"]["GNBeneficiaryRegionGeoCode"] = location["rgncd"]
+                except KeyError:
+                    try:
+                        # Wales, Scotland, Northern Ireland don't have rgnm so we set them as country
+                        grant["additional_data"]["GNBeneficiaryRegionName"] = location["ctrynm"]
+                        grant["additional_data"]["GNBeneficiaryRegionGeoCode"] = location["ctrycd"]
+                    except KeyError:
+                        pass
+
+        # recipientOrganisationLocation
```
rhiaro:
```suggestion
        # recipientOrganizationLocation
```

just for consistency..

rhiaro (dismissed):

dataload/import_to_elasticsearch.py
michaelwood: Yeah that should be covered by the earlier check on grant import, if it doesn't exist it will create an empty `dict`

https://github.com/ThreeSixtyGiving/grantnav/blob/live/dataload/import_to_elasticsearch.py#L420

PR #1103, opened 2024-09-03
docker: update docker setup

rhiaro:
This lets me run the tests and the django app locally. It doesn't use/include the provenance data. It's not perfect/complete, but if it doesn't make someone else's life more difficult, can we merge it in the spirit of incremental improvements?


michaelwood (approved):
Not tested but as you mentioned this is an improvement that you have checked and is not an application change then all good

PR #1116, opened 2024-10-18
templates: ThreeSixtyGiving -> 360Giving

### LINKED ISSUE #1111: Use 360Giving instead of threesixtygiving in copy referring to the organisation
The Additional Data section of grants pages says "Additional data added by ThreeSixtyGiving"

This has been like this for some time but it's a problem for us as it exacerbates the problem created by our domain of people thinking we are called ThreeSixtyGiving.

1. Please can this be changed to 360Giving on all grants pages?
2. Is there a way to find out whether this is used elsewhere in copy in our tools? I tried doing a search but it threw up too many false positives as we list our contact email grantnav@threesixtygiving.org in the footer, which means every page gets flagged.

michaelwood:
Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1111

### DIFF
--- a/grantnav/frontend/templates/grant.html
+++ b/grantnav/frontend/templates/grant.html
@@ -147,7 +147,7 @@
 
     <div class="spacer-4" style="display:none" id="additional-data-section">
       <div class="panel-default bottom-space">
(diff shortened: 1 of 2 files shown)


michaelwood:
requires #1118

PR #1118, opened 2024-10-18
docs: base: Remove non-functional theme option

### LINKED ISSUE #1117: helpsite "Uncaught ReferenceError: SphinxRtdTheme is not defined"
The helpsite is currently reporting an error trying to set "StickyNav" to false.

Unclear as to when that stopped being able to be accessed/set or if this is an upstream issue.

```html
  <script type="text/javascript">
      jQuery(function () {
          SphinxRtdTheme.StickyNav.enable(false);
      });
  </script>
```
docs/_templates/base.html:175

Which appears to be from the RTD theme : 

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L202

michaelwood:
This appears to access an out of scope variable and is actually no longer needed.

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1117

### DIFF
--- a/docs/_templates/base.html
+++ b/docs/_templates/base.html
@@ -170,11 +170,6 @@
       </footer>
     </div>
 
-  <script type="text/javascript">
-      jQuery(function () {
-          SphinxRtdTheme.StickyNav.enable({{ 'true' if theme_sticky_navigation|tobool else 'false' }});
-      });
-  </script>
   {% block cookieconsent %}
   {% include '_parts/cookie-consent.html' %}
   {% endblock %}


michaelwood:
Reviewer note: test version https://help.grantnav.threesixtygiving.org/en/mw-fix_helpsite_error/

michaelwood:
This will fix the test that is currently failing but only once deployed as it is referencing live

R2ZER0 (approved):
Looks good :+1:

PR #1119, opened 2024-10-29
org_utils: Prefer datastore-specified org ids list

R2ZER0:
If the loaded data package contains `non_primary_org_ids`, prefer it over `ftcData.orgIDs`.


michaelwood (commented):
Do we actually need to maintain backward compat? If the deploys are in the right order it should be fine? - I'm wary of adding some technical debt here that we won't have an opportunity or potential memory to remove in the future.

R2ZER0:
I've removed the backwards compatibility code path.

michaelwood (approved):

PR #1131, opened 2025-01-06
workflows: Pin to specific ubuntu version and remove actions cache

michaelwood:
ubuntu-latest is now changing and actions cache is no longer needed.


R2ZER0 (approved):
LGTM :+1:

PR #1138, opened 2025-01-21
requirements: Remove unused django-super-favicon

### LINKED ISSUE #1042: Remove django-super-favicon
I don't think this is useful anymore. The load template tag was added in 2016 before the existence of the 360 design system. django-super-favicon package brings in Pillow which often has security vulnerabilities. These are only used in the management commands so the risk is very low but it would be useful to reduce the security updates needed.

michaelwood:
Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1042

### DIFF
--- a/grantnav/frontend/templates/base.html
+++ b/grantnav/frontend/templates/base.html
@@ -1,5 +1,4 @@
 {% load static %}
-{% load favicon %}
 <!doctype html lang="en">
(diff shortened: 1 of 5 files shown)


R2ZER0 (approved):
Looks sensible :+1:

PR #1139, opened 2025-01-22
templates: org-publisher: Remove responsive table to fix overflow bug

### LINKED ISSUE #1079: Publisher section on org page over flow container
image

michaelwood:
The datatables is getting it's algorithms in a twist trying to fit all the content into the table. This behaviour isn't needed for this simple table so remove it.

Also fix the license to show the license_name instead of full URL to help with space use.

Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1079

### DIFF
--- a/grantnav/frontend/templates/components/org-page-publisher.html
+++ b/grantnav/frontend/templates/components/org-page-publisher.html
@@ -21,7 +21,7 @@
 
         <h4>Datasets</h4>
         <p>Here is an overview of grants published by {{publisher.name}}.</p>
-        <table class="table table--zebra table--small dt-responsive generic_datatables" style="width: 100%">
+        <table class="table table--zebra table--small" style="width: 100%">
         <thead>
             <tr>
             <th class="all">Dataset (download link)</th>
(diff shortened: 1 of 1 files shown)


michaelwood:
Turns out to be quite a reasonable amount of work to get datatables library to actually behave. I think there may be a conflict between the responsive design rules that are in the design system and the responsive design rules which are in the datatables library. This may also explain why we're seeing strange behaviour elsewhere.

Current live:

image

Proposed fix:

**Note** this removes sorting / search/ pagination features from this table (which I don't think are needed) 

image

michaelwood:
Another option would be to rewrite the datasets table and use a custom set of settings instead of the generic one, this is a bit more involved though so I thought worth seeing if this is acceptable.

mariongalley:
> Note this removes sorting / search/ pagination features from this table (which I don't think are needed)

Yes that's fine, most publishers don't have enough files to need pagination

mariongalley (approved):

R2ZER0 (approved):
Code changes look fine :+1:

PR #1147, opened 2025-03-03
templates: Update all instances of urls to go to new website structure

michaelwood:


codemacabre (changes requested):
Looks good, although there are a few links with a `title` attribute which seem superfluous. According to WCAG, links should only contain `title` attributes to provide [supplementary information](https://www.w3.org/WAI/WCAG21/Techniques/html/H33.html). Duplication of text in the link text means anyone using a screen reader will hear the text twice, which isn't good practice.

I recommend reviewing the `title` texts and deciding whether they're providing any useful supplementary information.

michaelwood:
> Looks good, although there are a few links with a `title` attribute which seem superfluous. According to WCAG, links should only contain `title` attributes to provide [supplementary information](https://www.w3.org/WAI/WCAG21/Techniques/html/H33.html). Duplication of text in the link text means anyone using a screen reader will hear the text twice, which isn't good practice.
> 
> I recommend reviewing the `title` texts and deciding whether they're providing any useful supplementary information.

Thanks, my rationale was that product names aren't descriptive so my hope was to add accessibility to those links.

codemacabre (dismissed):

grantnav/frontend/templates/datasets.html
mariongalley, line 1: We're not intending to publicly share the link to the unstyled Registry 'classic'. The options are to either refer to the Registry without providing a link (it's now an internal tool), which is the approach we've taken in the Publisher Guidance, or provide the link to the Quality Dashboard where signposting to a link of all publishers. I think in this case we should do the former.

grantnav/frontend/templates/grant.html
mariongalley, line 1: Correct link for negative/zero values guidance is https://www.360giving.org/explore/before-you-start/what-to-look-for/#negative-zero

grantnav/frontend/templates/home.html
mariongalley, line 1: Best link for GrantNav workshops is https://www.360giving.org/support/?event_type%5B%5D=intro-to-grantnav&sort-by=date

mariongalley (commented):
Added 3 comments on specific files - hopefully this process makes them discoverable

codemacabre (approved):

PR #1148, opened 2025-03-03
Apply a11y improvements

codemacabre:
- Add `h1` headings to all pages (hidden with the `screen-reader-only` CSS class where appropriate)
- Fix `select` elements accessible labels

### DIFF
--- a/grantnav/frontend/templates/components/dropdown-filter.html
+++ b/grantnav/frontend/templates/components/dropdown-filter.html
@@ -2,9 +2,9 @@
 <form id="sort-form" class="form-inline">
   <div class="dropdown-filter">
     {% if title %}
-    <label class="dropdown-filter__label" for="results-sort-dropdown">{{ title }}</label>
+    <label class="dropdown-filter__label" for="sort_options">{{ title }}</label>
     {% else %}
-    <label class="dropdown-filter__label" for="results-sort-dropdown">Sort By:</label>
+    <label class="dropdown-filter__label" for="sort_options">Sort By:</label>
     {% endif %}
       <select id="sort_options" name="sort" class="form-control" onchange="sort_changed(this.value)">
       {% with current_sort=query|get_current_sort %}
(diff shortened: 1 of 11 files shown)


grantnav/frontend/templates/components/dropdown-filter.html
```diff
@@ -2,9 +2,9 @@
 <form id="sort-form" class="form-inline">
   <div class="dropdown-filter">
     {% if title %}
-    <label class="dropdown-filter__label" for="results-sort-dropdown">{{ title }}</label>
+    <label class="dropdown-filter__label" for="sort-options">{{ title }}</label>
     {% else %}
-    <label class="dropdown-filter__label" for="results-sort-dropdown">Sort By:</label>
+    <label class="dropdown-filter__label" for="sort-options">Sort By:</label>
     {% endif %}
       <select id="sort_options" name="sort" class="form-control" onchange="sort_changed(this.value)">
```
michaelwood, line 9: The input id is `sort_options` here rather than `sort-options`

grantnav/frontend/templates/search_widgets.html
```diff
@@ -88,8 +88,8 @@ <h2>Settings:</h2>
                 </div>
                 {% include 'components/dropdown-filter.html' with title="Default Sort Order:" required=True options=dropdownFilterOptions %}
                 <div class="dropdown-filter flex-centre">
-                  <label class="dropdown-filter__label" for="select-views">Recency:</label>
-                  <select class="dropdown-filter__select" name="views" id="select-recency" onchange="updateRecency()">
+                  <label class="dropdown-filter__label" for="select-recency">Recency:</label>
+                  <select class="dropdown-filter__select" name="recency" id="select-recency" onchange="updateRecency()">
```
michaelwood, line 93: This looks like it might change a form key ...but I think this just uses the `select`s value in the JS (updateRecency()) and isn't an actual submitted form so should be all fine. Shout if that isn't your understanding!

michaelwood (changes requested):
Great - Just one small change needed

michaelwood (approved):

PR #1149, opened 2025-03-04
Replace organization with organisation everywhere in the UI

### LINKED ISSUE #1141: OrganiSation vs OrganiZation consistency
**Describe the bug**
GrantNav filters use organiSation and organiZation inconsistently. This history of this is:

1. We originally wanted to change the Funding Organization Name, Type and Recipient Organization Name filters to use S rather than Z - I'm not sure exactly why as it's before my time but I believe this was deemed to be too complicated.
2. We agreed that wherever we could control it, we would use S rather than Z, so we use S in values of the filters e.g. Grantmaking OrganiSation in the Funding OrganiZation filter, OrganiSation in the Recipient Type filter.
3. We've now reviewed this decision and feel it's better to be consistent from a user experience point of view.

**Expected behavior**
Our preference would be to use S in all places including the filter labels. Is it possible to change this?

If not, please can we change the filter values to use Z to match with the filter labels?

**Screenshots**

Image

Image

codemacabre:
Fixes https://github.com/ThreeSixtyGiving/grantnav/issues/1141

Note: these changes only affect instances of 'organization' in the UI, not variable names in the code itself or query parameters.

### DIFF
--- a/grantnav/frontend/templates/components/filters/groups/funder-filters.html
+++ b/grantnav/frontend/templates/components/filters/groups/funder-filters.html
@@ -5,7 +5,7 @@
       {% include 'tokens/accordion-toggle-icon.html' %}
     </summary>
     <div class="filter-group_contents-wrapper">
-        {% include 'components/filters/lists/filter-list-generic-checkboxes.html' with open=True aggregate_data=results.aggregations.fundingOrganizationTSGType filter_title="Funding Organization Type" %}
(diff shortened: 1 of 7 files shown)


R2ZER0 (approved):
Does what it says on the tin :+1:

mariongalley:
@codemacabre Thanks for working on this - is it possible that this hasn't percolated to live yet? I still see Zs in the filters

image

michaelwood:
Now deployed

Issue #801, opened 2021-12-01
Version information incorrect


michaelwood:
In the footer GrantNav reports:
`Running version pre-2020-sprint-233-g6f45a2b.` 

This is incorrect, not sure why. It might be simpler to not use the library which is generating this and use something similar to in the datastore: https://github.com/ThreeSixtyGiving/datastore/blob/0a2798444b92f74a156c0b6e07a85dc117814274/datastore/ui/views.py#L91

BibianaC:
![](https://github.trello.services/images/mini-trello-icon.png) [Version information incorrect](https://trello.com/c/3zX8BLCk/240-version-information-incorrect)

mariongalley:
Remove version

michaelwood:
Steps for done are:

1) Remove the grantnav version from the footer: `Running version pre-aggregates-api-136-g2d3c599`

![Image](https://github.com/user-attachments/assets/bf0a2855-df45-4f05-9778-abb28cfb56f7)
2) Remove the code that creates the `request.tag` 

Add the git version information to the about.html page https://grantnav.threesixtygiving.org/about:

3) In urls.py edit about.html url to use the context_extra={} in the as_view() function
4) Make a util similar to https://github.com/ThreeSixtyGiving/datastore/blob/live/datastore/ui/views.py#L91
5) Add version information to the about page

BibianaC:
Done

PR #1154, opened 2025-03-12
grantnav: frontend: Add the git version information to the about.html

BibianaC:
Steps 3, 4 & 5 of #801

### DIFF
--- a/grantnav/frontend/templates/about.html
+++ b/grantnav/frontend/templates/about.html
@@ -16,6 +16,7 @@
         <p>{% blocktrans %}For more information about GrantNav and the data used in it, see the links in the footer below.{% endblocktrans %}</p>
         <p style="font-style: italic">{% blocktrans %}GrantNav was developed for 360Giving by Open Data Services. The project received funding from the Esmée Fairbairn Foundation, Indigo Trust and Nesta.{% endblocktrans %}</p>
         <p style="font-style: italic">{% blocktrans %}We welcome feedback and questions on GrantNav. Please <a href="https://www.360giving.org/contact/" title="Contact 360Giving">contact us</a>{% endblocktrans %}</p>
+        <p style="font-style: italic">{% blocktrans %}Current Git revision: {{ git_revision }} {% endblocktrans %}</p>
       </section>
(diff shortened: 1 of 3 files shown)


michaelwood:
Don't forget to also update requirements_dev, that was missing from https://github.com/ThreeSixtyGiving/grantnav/pull/1151 thanks

BibianaC:
> Don't forget to also update requirements_dev, that was missing from #1151 thanks

I changed the requirements, not requirements_dev. Should I also change requirements_dev even if I haven't made any changes? @michaelwood

R2ZER0:
@BibianaC Yes, please re-compile dev requirements too :)

dev requirements include requirements ( https://github.com/ThreeSixtyGiving/grantnav/blob/live/requirements_dev.in#L1 ), so any changes to requirements will also change dev requirements.

BibianaC:
Thanks @R2ZER0 . I have re-compiled.

requirements_dev.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.12
```
R2ZER0: @BibianaC Sorry last thing - please could you re-compile using Python 3.8 instead of Python 3.12 , so it's the same as `requirements.txt` https://github.com/ThreeSixtyGiving/grantnav/blob/live/requirements.txt#L2
 ... we'll get around to upgrading python one day soon :)

BibianaC: @R2ZER0 sorry, I didn't realise about the version. Now it is with version 3.8.

michaelwood (changes requested):
I re-based your branch and squashed the unnecessary commits however the git version is always "unknown", could you re-test this please and see if there is an issue.

BibianaC:
@michaelwood @R2ZER0 it should be working now.

michaelwood:
> @michaelwood @R2ZER0 it should be working now.

Great thanks

michaelwood (approved):

PR #1157, opened 2025-03-17
Add global nav bar

codemacabre:
Add global nav bar from design system:

image

image

image


michaelwood (dismissed):
All good from a code point of view. :+1:

mariongalley (dismissed):

michaelwood (approved):

PR #1158, opened 2025-03-17
Add loading overlay for when filters change

### LINKED ISSUE #851: Create a loading overlay screen when selecting different filter options
When different filter options are selected you can select multiple filter items but only one will apply due to the fact that selecting a filter option causes a page reload.

Add a loading screen to avoid people hitting the race condition of selecting multiple things.

codemacabre:
- Adds `show_overlay` function to display the loading overlay when filters are changed
- Adds event listeners to `filter-list__filter-item` links to ensure the overlay is displayed when they are clicked / touched / 'Enter' is pressed

Fixes #851

### DIFF
--- a/grantnav/frontend/templates/search.html
+++ b/grantnav/frontend/templates/search.html
@@ -9,7 +9,7 @@
 
 {% block main_content %}
   <main class="layout__content">
-
+    {% include 'components/loading-overlay.html' %}
     <div class="grantnav-search__wrapper">
       <div class="grantnav-search__sidebar">
         {% if results and results.hits.hits %}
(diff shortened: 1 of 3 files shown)


grantnav/frontend/templates/search.html
```diff
@@ -164,6 +164,21 @@ <h1>Search input is not valid</h1>
     $('input[type=radio][name=results-switcher][value="card"]').prop('checked', true);
   }
 
+  // Show overlay when any of the filter links are selected
+  const filters = document.querySelectorAll('.filter-list__filter-item');
```
michaelwood: I think this selector might be too broad? - the radio buttons for the include/exclude are in-page actions so don't need to trigger the loading overlay.

michaelwood (changes requested):
Need to check on the radio buttons on the filters, we might need to be more explicit about the selector, not quite sure.

codemacabre:
I've refined the selector slightly, but the radio buttons were triggering the overlay because it was being called too early in the `maybe_redirect_on_exclude`function, also resolved.

michaelwood (approved):

Issue #1122, opened 2024-11-11
Improve error on custom search endpoint with no params


michaelwood:
If someone / a bot hits `search_custom.csv` end point without specifying any parameters we generate an error report as this isn't a handled condition. 

Dev notes:

The existing error handling isn't working due to a typo in 
https://github.com/ThreeSixtyGiving/grantnav/blob/live/grantnav/frontend/user_csv_layout.py#L16

```diff
-fields = json.loads(request.POST.get("selection", []))
+fields = json.loads(request.POST.get("selection", "[]"))
```

BibianaC:
@michaelwood @R2ZER0 I have fixed the typo https://github.com/ThreeSixtyGiving/grantnav/pull/1159. This issue is to improve the error, how would you like this to be done?

michaelwood:
I think that should change in #1159 should do the trick, thanks @BibianaC

BibianaC:
Done

PR #1159, opened 2025-03-18
frontend: user_csv_layout: fix error on custom search endpoint with no params

BibianaC:
Related to #1122


michaelwood (approved):
Thanks the changes in on line 18 look good. For future please do not re-order the imports in this commit, the change isn't related and could result in a human error problem being missed in a code review such as accidentally removing an import.

PR #1162, opened 2025-03-18
tests: Re-enable running browser tests in Firefox

R2ZER0:
This commit fixes the initialisation code for Firefox GeckoDriver, re-enabling running browser tests in Firefox by setting env var BROWSER=Firefox or FirefoxHeadless.
While the tests run, they don't pass due to the current method used to fetch browser console logs.


michaelwood (approved):
Looks good, I don't have a setup to test this but trust you've done so!

PR #1168, opened 2025-03-21
frontend: templates: Add ids and anchor hashes to all filters

### LINKED ISSUE #961: GrantNav search & widget builder - don't go to top of page after applying a filter
**Is your feature request related to a problem? Please describe.**
When I'm applying filters, GrantNav jumps to the top of the search page every time, which makes it frustrating to have to scroll back down to apply more filters

**Describe the solution you'd like**
Remember where the user is when applying the filter and take them back on reloading the results.

codemacabre:
Fixes #961

Ids are added to filter parent details elements, then all anchors and javascript-triggered page reloads are injected with that id as an anchor hash.

### DIFF
--- a/grantnav/frontend/templates/components/filters/lists/filter-list-amount-awarded.html
+++ b/grantnav/frontend/templates/components/filters/lists/filter-list-amount-awarded.html
@@ -1,13 +1,13 @@
 {% load get_amount_range get_amount from frontend %}
 <div class="filter-list margin-bottom:2">
-  <details class="filter-list__accordion" open>
+  <details id="filter-accordion-amount-awarded" class="filter-list__accordion" open>
     <summary class="filter-list__label">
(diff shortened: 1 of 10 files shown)


michaelwood (commented):
There is a strange thing here where when the page loads it loads at the top and then scrolls to the content, is this something to do with the redirect? iirc a hash anchor should be able to cause the page to be rendered at the desired position.

codemacabre:
@michaelwood I believe most browsers load pages with the scroll position at the top first so the rest of the DOM can load [before scrolling to the hash position](https://www.w3.org/TR/2011/WD-html5-20110113/history.html#scroll-to-fragid) (at least that's what I understand from the spec). It's more noticeable here because the page is fairly complex so it takes a few milliseconds more to load the DOM before updating the scroll position.

We could add the CSS `scroll-behavior: smooth;` to the HTML to make it scroll more naturally rather than the harsher snap to position, or we could hide the page content until the DOM has loaded, which will cause a flash of empty content but might be a better option than the animated scroll.

michaelwood (approved):

michaelwood:
Thanks now deployed

Issue #1011, opened 2023-10-04
Use parallel threads to load grant data into GrantNav


michaelwood:
Reloading the GrantNav index is now taking about 2 hours to load, this mostly impacts development as any changes to the elasticsearch index require the data to be reloaded. I think it's now at a threshold that would warrant some time invested in speeding this up with some parallel threading on the importer module.

michaelwood:
Because we don't know how big each file will be we may need to chop up the loaded data into segments e.g. of 500 grants.

michaelwood:
Grant Loading code: https://github.com/ThreeSixtyGiving/grantnav/blob/live/dataload/import_to_elasticsearch.py#L388 , full data packages can be obtained from https://store.data.threesixtygiving.org/grantnav_packages/

michaelwood:
https://elasticsearch-py.readthedocs.io/en/latest/helpers.html

radix0000:
So, what sort of speed are we aiming for here and what is the typical spec of machines where loading is run, and where performance matters (are we mainly talking about devs laptops, on servers - don't have access to 360 servers)? Have got loading all data down to ~25 mins on my laptop, but this is a diminishing returns sort of situation (further improvements will take more and more time for less and less gain), so don't want to over-optimise if that is good enough for now.

So far have done some profiling and identified the two main performance issues as parsing JSON (about 2/3 of runtime) and the fact that ES org lookups are taking place during indexing of grants (a large chunk of rest of time). Note that loading data off disk, and writing to ES seem to be small compared to these issues, so concurrency (asyncio) isn't going to help. Did briefly explore threading but as expected doesn't seem to help as main performance issues are in Python code where locking is going to stop any parallel execution.

Ended up doing the following:

1. Refactoring `grant_generator` so that code to parse and process a file is in one function call
2. Spinning up a process pool (with process for each core - note speedup is going to depend on how many cores are available) using `multiprocessing` to handle processing files in parallel, and farm out individual files to that pool, which feeds back grants via queue to `grant_generator`
3. Cache orgs so that no ES lookups happen during grant indexing, not using the existing semi-caching mechanism, but filling a cache with all orgs as they are indexed (so zero lookups happen later), and only placing org data in cache that is required during grant indexing (so as not to use large amounts of memory)
4. Also  have unset ES refresh_interval before indexing, and set again afterwards (though unclear if this is having a significant effect on performance)

[...]

PR #1173, opened 2025-03-28
import_to_elasticsearch.py: Improve performance

michaelwood:
Refactor, add process pool, org caching and ES tuning https://github.com/ThreeSixtyGiving/grantnav/issues/1011


michaelwood:
We might need to use something like get_totals and poll that until it reaches the expected total? https://github.com/ThreeSixtyGiving/grantnav/blob/live/grantnav/frontend/views.py#L441

michaelwood:
```
=====================================================================
FAIL: test_links (tests.tests_links.LinkCheckTests)
Load each path and check the links within the page respond HTTP success
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/michael/dev/ods/threesixtygiving/grantnav/tests/tests_links.py", line 104, in test_links
    check_page_for_broken_links(page)
  File "/home/michael/dev/ods/threesixtygiving/grantnav/tests/tests_links.py", line 89, in check_page_for_broken_links
    assert not broken, f"Links broken on page {path}: {errors}"
AssertionError: Links broken on page /search: http://localhost:48739/org/GB-COH-02273708 (404)
```
This issue is actually present in the current live, however I think the data loading order is slightly changed with this performance improvement, which in turn means that on the first page of the search results we are seeing a grant which the orgs data comes via a grant which isn't loaded in the test data so even before this PR that org-id would have 404'ed. 

I think the fix for this is to find the right test data to load to make sure the org-id has a valid org.. will look into that.

R2ZER0 (approved):

PR #1175, opened 2025-04-10
ci: deps: Upgrade to Python 3.12

R2ZER0:


.tool-versions
michaelwood, line 1: How do we keep this up to date? That particular version info seems like it is look-up-able from various places as the SSOT so would be cautious about manually replicating that info

michaelwood (changes requested):
Looks good to me - just one query

PR #1204, opened 2025-09-25
Mw/countries filter v1

michaelwood:
This work adds countries filters to grantnav, re-works the results cards design to better explain the location data and removes UK assumptions from labels such as "district"


michaelwood:
Adding @mariongalley for final sign off from 360

mariongalley (dismissed):
We're good to go!

jarofgreen (approved):

PR #1208, opened 2025-10-06
Mw/fix funder sort order by

michaelwood:
Fixes sorting by funder. Copy change TBC `Total GBP Amount - Highest First"` to `Total GBP Amount to Organisations`. 

image

and 

image

Requires a data reload.


mariongalley (dismissed):
Happy with the copy change. Haven't had the opportunity to test but if you're satisfied the sort order works and no regressions have been introduced then I'm OK to approve.

michaelwood:
Bypassing rules - as no one available for code review and this is currently broken in live.

PR #1211, opened 2025-10-08
views: search: Default order latest grants by date unless text search

### LINKED ISSUE #1193: Best match default for keyword search only, otherwise default to latest date
**Is your feature request related to a problem? Please describe.**
Now that we have increased the number of useful filters, people are more likely to search using filters and may not be using keywords. However, Best Match is completely meaningless when no filters are applied, and the results are often grants from many years ago which are less relevant.

**Describe the solution you'd like**
Either GrantNav should default to latest date but then automatically apply best match when keywords are entered, or we should see if we can change the 'best match' logic to prioritise recency when there are no keywords.

**Describe alternatives you've considered**
We could exclude older grants from GrantNav, or pre-filter the main search to exclude older grants but still allow people to change the filter.

**Additional context**
Currently the empty search is showing grants ranging from 2015 to 2018 for me on the first page. I just filtered to the National Lottery Heritage Fund as an example and the entire first page of results is from 2005 (screenshot below):

Image

michaelwood:
Fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1193


michaelwood:
Now available to test on dev.

Tests cases are broken because of reliance of search order, will fix once intended sort order behaviour is confirmed.

mariongalley:
@michaelwood Had a go at testing on dev and couldn't get it to load - this is the /search page

image

michaelwood:
> @michaelwood Had a go at testing on dev and couldn't get it to load - this is the /search page
> 
@mariongalley 

I've reloaded dev (browser cache clear is needed (shift refresh)) and in doing so found a few more cases that needed addressing.

default = award date latest first

**home page**
* When no text in the search field do not change the ordering from the default, check the text input before submitting the search, if empty avoid setting up a "chosen" order-by within the search by redirecting to an empty search results page instead of submitting the form.

**search results page search**
* On an empty search revert to the default ordering
* If a text search is run and then the user delete the contents of the text search unset the best match ordering back to default
* If the search order has been changed by the user after a text search maintain that ordering 
* If an empty search is filtered on with the default ordering and then a text search is added don't override the ordering 
* If the user inputs a text search and then filters do not change ordering from best match
* If the user does a text search and applies a filter, then removes the text search don't reset the search

BibianaC (approved):

michaelwood:
Note for self re-check if there is something we can do about:

1. Apply Filter "London": sort order is Date
2. Add text "garden": sort switches to Best match
3. Add filter "To individuals": sort stays best match
4. Remove text "garden": sort stay best match - here I think it ought to switch back to date

Expected - revert back to awardDate

PR #1219, opened 2025-10-22
import_to_elasticsearch: Ensure ALL orgs are cached not just canonical

michaelwood:
When caching the org it's important to make sure the cache key (org-id) also has values for all the other known org-ids for that organisation.

e.g.

cache[org-id-1] = {"name": "example org", "orgId": "org-id-1" ...}
cache[org-id-2] = {"name": "example org", "orgId": "org-id-1" ...}

Otherwise the mechanism to de-duplicate the organisations stops functioning.


michaelwood:
e.g. This PR | Live 

image

This regression was caused by the dataloader performance improvements with caching being to naive

BibianaC (approved):

PR #1220, opened 2025-10-23
Add countries include/exclude type filters

michaelwood:


michaelwood:
@mariongalley - Just to double check before I do the wiring up of this:

This type of filter for best/grant location/recipient org countries:

image

mariongalley:
@michaelwood That's the one, thanks!

mariongalley:
Can/should we do this for the other location filters too?

michaelwood:
> Can/should we do this for the other location filters too?

The main negative for this kind of filter is that it doesn't surface the information to the user until they start interacting with it (and it's a bit slow). 

We also have sufficient number of these widgets that after these 3 are implemented it will be needed to refactor the way they work to reduce the code complexity. Happy to do just will be a day needed instead of hours.

mariongalley:
@michaelwood Discussed with Ruth and agreed the other filters don't need it. The countries & regions filters has a small enough number of values that you could tick all but one by hand. The others don't really have a clear use case for all but one county or all but one LA.

BibianaC (dismissed):

codemacabre (dismissed):
LGTM!

BibianaC (approved):

PR #1223, opened 2025-10-31
templatetags: Remove sort on field names in flatten process

### LINKED ISSUE #1222: Group location fields by object rather than field name in the GrantNav grant pages
Currently, where multiple locations are published for a grant, the grant page will show all the details of these locations grouped by the field names eg. [](https://grantnav.threesixtygiving.org/grant/360G-SE-2024018693_2025-03-31) Eg. it shows five Beneficiary Location: Geographic Code's  then five Beneficiary Location: Geographic Code Type's and so on - same in the additional_data: locationLookup fields.  

It would make it more understandable and easy to read if they were grouped by each location object rather than the field names so that the related fields are all grouped together for each location record.

michaelwood:
fixes: https://github.com/ThreeSixtyGiving/grantnav/issues/1222

### DIFF
--- a/grantnav/frontend/templatetags/frontend.py
+++ b/grantnav/frontend/templatetags/frontend.py
@@ -84,8 +84,8 @@ SKIP_KEYS = ["Identifier", "Title", "Description", "filename",
 
 @register.filter(name='flatten')
 def flatten(d):
-    return sorted([(key, value) for key, value in flatten_dict(d)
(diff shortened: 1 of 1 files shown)


BibianaC (approved):

R2ZER0 (approved):

michaelwood:
Deployed

Issue #858, opened 2022-08-16
Search results CSV download not in same order as on page


drkane:
**Describe the bug**

When downloading search results from GrantNav, the resulting CSV is in a different order to the results from the search page.

**To Reproduce**

1. Run a search on GrantNav
2. Look at the top search result
3. Download the search result
4. Top search result is not the top row in the spreadsheet

**Expected behavior**

Expected the CSV results to reflect the search order found in the HTML page

**Additional context**

Based on working through search results with a user - the behaviour was surprising for them, and stopped them from doing what they wanted to do (which was to check the last search results manually as they had the least confidence in them)

mariongalley:
Fixing this would fix #351

michaelwood:
This requires the order-by to be preserved e.g. Amount - lowest first - https://grantnav.threesixtygiving.org/search?query=*&default_field=*&sort=amountAwarded+asc&recipientRegionName=England

BibianaC:
@mariongalley would you implement this fix for all the CSV downloads, or as it was proposed in #351, only for result sets < 10k?

mariongalley:
@BibianaC Unless it's substantially more effort I would do it for all downloads

BibianaC:
@ruth360Giving, this is now deployed if you want to test it.

mariongalley:
I've had a quick test and it looks good, thanks @BibianaC (FYI @ruth360Giving )

Issue #351, opened 2016-11-03
Make sure sort order of CSV or JSON data matches view user is looking at on results page


TomSteinberg:
This may be a nit-pick, but please see it as a sign of my usage of the site :)

After carefully constructing a query with all the right keywords and modifiers, and sorting it exactly like I want, I hit CSV to export that data for my own use. But when the CSV opens in Excel the sort order is not the same as the one I was looking at before (i.e the data is not arrange by grant size from largest to smallest). It's just a small thing but it might confuse some people if the spreadsheet they open appears to be different from the web page they were just staring at.

kindly:
That is a fair nit-pick.  I am not sure there is an easy solution to it though..

It is due to a bit of a compromise we had to make for making the download functionality fast and not take up too much server resources.

On the back-end we use elasticsearch to do all the searching and sorting.  It has a feature that allows you to stream the data quickly for large amounts of data and not take up too much server resources i.e you can download all the grants if you want.
The feature is called scan and documented here http://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.scan
Scan has an option called preserve_order and it recommends you **not use** it as it negates the speed and memory usage benefits of using that elastic search feature.  So we decided to not use that option for that reason.

That being said, we have not tested it with the option on, it may not make that much difference in our case and our server may be fine with it.
  
The other option would be to only switch it on for "small" result sets say under 10k grants or so as I am sure that would not be too bad on the server.  That fix would be a bit more complicated and may be confusing for the user that sometimes it works and sometimes not.

timgdavies:
One possibility would be to use this as a 'learning moment' for data users.

For example:

- For result sets < 10k perhaps turn the option on by default
- For larger results, keep it off - and generate a message to users that
appears after download is selected (i.e. start download, but pop-up a
modal)

We could have a modal after downloads anyway which provides some basic
data-use tips such as:

- Turning on filters in your spreadsheet to sort and filter the data;
- Creating a pivot from the spreadsheet;

And which adds in the note that 'because there were more than 10,000 rows,
we've not sorted things on the server - but you can do that locally'.

TomSteinberg:
So, as I said, this is a nit-pick not a show-stopper. Please prioritise accordingly!

mariongalley:
Fixed for CSV in #858

PR #1246, opened 2026-03-30
frontend: views: preserve order in CSV downloads

BibianaC:
Relates to https://github.com/ThreeSixtyGiving/grantnav/issues/858 and https://github.com/ThreeSixtyGiving/grantnav/issues/351

Screenshot 2026-03-30 at 09 08 54

Screenshot 2026-03-30 at 09 09 20

### DIFF
--- a/grantnav/frontend/views.py
+++ b/grantnav/frontend/views.py
@@ -145,7 +145,7 @@ SEARCH_SUMMARY_AGGREGATES = {
 def grants_csv_generator(query, grant_csv_titles, grant_csv_paths):
     yield grant_csv_titles
     es = get_es()
-    for result in scan(es, query, index=get_index()):
+    for result in scan(es, query, index=get_index(), preserve_order=True):
         result_with_provenance = {
             "result": result["_source"],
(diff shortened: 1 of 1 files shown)


mrshll1001 (approved):
I'm happy to approve this based on my understanding of the change and Elastic Search's `scan` API. I think this addresses the problems raised in the issues.

We should note though that we may see a performance hit by using `preserve_order=True`, as cited by the ElasticSearch docs:

* https://www.elastic.co/docs/reference/elasticsearch/clients/python/client-helpers#scan

I'm not sure whether the performance hit is large or not, and I'm assuming that exporting CSV files is an adhoc use rather than something that someone is doing a lot of. I think it's fine but we should pay attention to this if issues are raised about performance in this area.

mariongalley:
Has this been deployed to production @BibianaC ? I just did a test in live and the sort order wasn't preserved

BibianaC:
@mariongalley this is not deployed yet. We will deploy this week.

mariongalley:
Thank you. When it's ready can you ask @ruth360Giving to test please?

Issue #1225, opened 2025-11-21
Number fields stored as text don't allow searching with 'TO' operator, '<' or '>'


mariongalley:
**Describe the bug**
Now that we have fixed the specific field search, we want to use the search bar for searching durations (other examples we tried were searching in recipient org income and planned dates). However, because these fields are stored as text, it means it's not possible to search for e.g for duration '12 TO 36' or duration '>12'.

**Expected behavior**
I expect to be able to search number fields using these characters and operators.

BibianaC:
@mariongalley, could you attach a screenshot of how you search, please?

ruth360Giving:
Hi - here are some examples:

[**additional_data.recipientOrgInfos.latestIncome: [10000 TO 100000]**](https://grantnav.threesixtygiving.org/search?query=additional_data.recipientOrgInfos.latestIncome%3A+%5B10000+TO+100000%5D&default_field=%2A&sort=_score+desc&_redirect_src=int)

<img width="1126" height="741" alt="Image" src="https://github.com/user-attachments/assets/121f44e4-dbaf-4bdb-be37-6356d4164c09" />

- Returns no results but should match the results on GrantVis

[**plannedDates.duration>12**](https://grantnav.threesixtygiving.org/search?query=plannedDates.duration%3E12&default_field=%2A&sort=_score+desc&_redirect_src=int)

<img width="1239" height="595" alt="Image" src="https://github.com/user-attachments/assets/4bde8c04-fa49-41c3-9711-a8437b781222" />

- returns inaccurate results where duration is not greater that 12

BibianaC:
Thanks @ruth360Giving . It might return results, but not be accurate.

mariongalley:
I'm not sure this is working correctly. I ran the 2 examples that Ruth shared as tests in dev and downloaded the results:
- additional_data.recipientOrgInfos.latestIncome: [10000 TO 100000] includes ~ 25 grants with Latest Recipient income < £10k and ~ 75 grants with Latest Recipient income > £100k.
- plannedDates.duration>12 includes all sorts including empty duration, 0 values and durations <12

PR #1251, opened 2026-05-15
Fix Number fields store as text

BibianaC:
Relates to #1225

### DIFF
--- a/tests/tests_numeric_field_search.py
+++ b/tests/tests_numeric_field_search.py
@@ -1,0 +1,161 @@
+import time
+import os
+
+from django.test import TestCase, Client
+
+from dataload.import_to_elasticsearch import import_to_elasticsearch
+
+prefix = os.path.join(os.path.dirname(__file__), "data")
+
+
+class NumericFieldSearchTest(TestCase):
+    """
+    Tests for searching numeric fields using range and comparison operators.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        import_to_elasticsearch(
+            [
+                os.path.join(prefix, "a002400000KeYdsAAF.json"),
+                os.path.join(prefix, "grantnav-20180903134856.json"),
+                os.path.join(prefix, "a002400000nO46WAAS.json"),
+            ],
+            clean=True,
+            funders=os.path.join(prefix, "funders.jsonl"),
+            recipients=os.path.join(prefix, "recipients.jsonl"),
+        )
+        time.sleep(5)
+
+    def setUp(self):
+        """
+        This setUp method will be executed before each test method.
+        """
+        self.client = Client()
+
+    def test_planned_dates_duration_range_query(self):
+        """
+        Test that range queries work on plannedDates.duration field
+        """
+
+        # Query for durations between 12 and 36 months
+        r = self.client.get('/search?query=plannedDates.duration%3A+%5B12+TO+36%5D', follow=True)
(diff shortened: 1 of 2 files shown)


BibianaC:
@chrisarridge The links check is failing because of the Standard.

The failing test is the aggregates test. It’s the same issue as the geolocation one. I’ve made some changes in the aggregates file. The test is passing locally for me, but it’s still failing in GitHub.

chrisarridge:
> @chrisarridge The links check is failing because of the Standard.
> 
> The failing test is the aggregates test. It’s the same issue as the geolocation one. I’ve made some changes in the aggregates file. The test is passing locally for me, but it’s still failing in GitHub.

The aggregates test fails for me locally as well.  Interestingly, if I drop both the commits that change `aggregates_expected.json` then the test passes absolutely fine.  Would you expect the aggregates to change for your change in this case?  I'm curious about this given that we are also having issues with the geolocation work.

BibianaC:
@chrisarridge the process is:
- I make changes.
- Test_api fails.
- I uncomment the commented code.
- Run tests.
- Test Api fails.
- Comment back the code and change the expected aggregates file.
- Run tests.
- Everything is ok.

BibianaC:
@chrisarridge I have added the file to live. Here, the tests are passing, but locally are not.

tests/tests_numeric_field_search.py
```diff
@@ -0,0 +1,126 @@
...
+        Test that range queries work on additional_data.recipientOrgInfos.latestIncome field
+        """
+
+        # Query for organisations with latest income between 10,000 and 500,000
+        r = self.client.get('/search?query=additional_data.recipientOrgInfos.latestIncome%3A+%5B10000+TO+500000%5D',
+                           follow=True)
+        self.assertEqual(r.status_code, 200)
+        total_hits = r.context["results"]["hits"]["total"]["value"]
+        self.assertGreater(total_hits, 0,
+            "Range query should return results for income between 10,000 and 500,000")
```
chrisarridge, line 149: Could we add something here that checks the `latestIncome` fields to make sure they match the range?  For example,

```
for hit in r.context["results"]["hits"]["hits"]:
    recipientOrgInfos = hit["_source"].get("additional_data",{}).get("recipientOrgInfos",[])
    for recipientOrg in recipientOrgInfos:
        if "latestIncome" in recipientOrg:
            self.assertLessEqual(int(recipientOrg["latestIncome"]), 500000)
            self.assertGreaterEqual(int(recipientOrg["latestIncome"]), 10000)
```

chrisarridge: In principle such a check could also be added to the income greater than, and the durations extending similarly.

chrisarridge (commented):
Looks good.  I have a suggestion for expanding the numeric search test.  Please could you squash the two dataload modification commits, and the three test data update commits to just leave two commits in the PR?  I think this might be cleaner and easier to follow in the repo history.

mariongalley:
Please can you spin up a dev GrantNav so we can test this? Changes to search functionality have historically had a lot of unintended side-effects and I wouldn't want to put this live without functional testing.

chrisarridge:
> Please can you spin up a dev GrantNav so we can test this? Changes to search functionality have historically had a lot of unintended side-effects and I wouldn't want to put this live without functional testing.

Yes @mariongalley we can do this.

BibianaC:
Thanks @chrisarridge . I have made the changes.

